PR #1, opened 2026-03-20
try fixing problem with CTAS when attaching

guillesd:
The problem was that CREATE TABLE AS uses a different `RpcInsert` constructor that sets table = nullptr (only schema + info are set). Both `GetGlobalSinkState` and `Sink` were blindly dereferencing table.

The fix in `GetGlobalSinkState`:
  - INSERT path (table is set): uses the existing table entry as before
  - CTAS path (table is null): sends a `CatalogRequestMessage` to create the table on the remote side, then constructs an `RpcTableCatalogEntry` owned by the global state, so Sink can use it for appending rows

This also fixed a latent bug in `RpcSchemaCatalogEntry::CreateTable` where the returned `unique_ptr` was being converted to `optional_ptr` and immediately destroyed (dangling pointer) — by inlining the same logic and keeping ownership in `RpcInsertGlobalState`.

### DIFF
--- a/src/rpc_insert.cpp
+++ b/src/rpc_insert.cpp
@@ -1,5 +1,6 @@
 #include "rpc_insert.hpp"
 
+#include "duckdb/catalog/catalog_transaction.hpp"
 #include "duckdb/planner/operator/logical_insert.hpp"
 #include "duckdb/planner/operator/logical_create_table.hpp"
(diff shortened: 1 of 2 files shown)


guillesd:
@hannes I'm not 100% sure moving table entry to the global state is the way, if you have a different pointer let me know!

PR #9, opened 2026-03-31
Fix concurrency bug

### LINKED ISSUE #8: http transport corrupts under concurrent connections
After ~8 concurrent connections, the HTTP server starts returning truncated
or corrupted binary responses, causing:
  1. `Failed to commit: Failed to send message Failed to read connection`
  2. `cannot start a transaction within a transaction` (stale transaction state)
  3. `Failed to deserialize: not enough data in buffer` (all subsequent ops)

Once a connection hits error #1, it never recovers — every subsequent query on
that connection fails with the deserialization error.

> This has only happened on the read connections for me

Unix socket transport handles the same concurrency level without errors as did the previous Web Socket implementation.

```python
import os
import time
from concurrent.futures import ThreadPoolExecutor, as_completed

import duckdb

EXTENSION_PATH = os.path.expanduser(
    "~/Dev/duckdb-rpc/build/release/extension/rpc/rpc.duckdb_extension"
)
LISTEN = "http://localhost:19444"
NUM_READERS = 8
NUM_WRITERS = 4
ITERATIONS = 50


def main():
    # Start server
    server = duckdb.connect(config={"allow_unsigned_extensions": "true"})
    server.execute(f"LOAD '{EXTENSION_PATH}'")
    server.execute("CREATE TABLE test_data AS SELECT i, 'row_' || i AS name FROM range(1000) t(i)")
    server.execute(f"CALL rpc_start('{LISTEN}')")
    time.sleep(0.2)

    # Shared client connection — workers use .cursor()
    client = duckdb.connect(config={"allow_unsigned_extensions": "true"})
    client.execute(f"LOAD '{EXTENSION_PATH}'")

    def reader(worker_id):
        cur = client.cursor()
        cur.execute(f"ATTACH '{LISTEN}' AS r{worker_id} (TYPE rpc)")
        errors = []
        for i in range(ITERATIONS):
            try:
                cur.execute(f"SELECT count(*) FROM r{worker_id}.main.test_data WHERE i > {i * 10}").fetchall()
            except Exception as e:
                errors.append(str(e).split("\n")[0])
        cur.close()
        return "reader", worker_id, errors

    def writer(worker_id):
        cur = client.cursor()
        cur.execute(f"ATTACH '{LISTEN}' AS w{worker_id} (TYPE rpc, READ_WRITE)")
        errors = []
        for i in range(ITERATIONS):
            row_id = 100_000 + worker_id * ITERATIONS + i
            try:
                cur.execute(f"INSERT INTO w{worker_id}.main.test_data VALUES ({row_id}, 'written_{row_id}')")
            except Exception as e:
                errors.append(str(e).split("\n")[0])
        cur.close()
        return "writer", worker_id, errors

    # Run concurrent readers + writers
    with ThreadPoolExecutor(max_workers=NUM_READERS + NUM_WRITERS) as pool:
        futures = []
        for i in range(NUM_READERS):
            futures.append(pool.submit(reader, i))
        for i in range(NUM_WRITERS):
            futures.append(pool.submit(writer, i))
        for f in as_completed(futures):
            role, wid, errs = f.result()
            if errs:
                print(f"{role}-{wid}: {len(errs)} errors")
                for e in dict.fromkeys(errs):
                    print(f"  - {e}")
            else:
                print(f"{role}-{wid}: OK")

    # Cleanup
    client.close()
    server.execute(f"CALL rpc_stop('{LISTEN}')")
    server.close()


if __name__ == "__main__":
    main()
```

guillesd:
fixes #8 

### Root Cause 

The server's ThreadPool(8, 64) created only 8 fixed threads (the second parameter is max queue size, not max threads). In cpp-httplib, each keep-alive HTTP connection holds a server thread for its entire lifetime (blocking on socket reads between requests). With 8+ concurrent ATTACHes, all 8 threads get occupied by catalog client keep-alive connections. When the scan function then creates a new RpcClient to FETCH data, it can't get a server thread — they're all blocked waiting for the next request from the catalog clients, which in turn are waiting for the scan to complete. Deadlock.

There was also a problem with multiple threads being inside of MakeRequest leading to this pattern:
```
  Thread A: Send(msg_A)     → stores result_A in https_result
  Thread B: Send(msg_B)     → overwrites https_result with result_B
  Thread A: Receive()       → reads result_B instead of result_A  ← WRONG
```
### Changes:
1. src/https_rpc_server.cpp: Changed ThreadPool(8, 64) to ThreadPool(128). With 128 threads, there's capacity for catalog connections, scan thread connections, and new incoming connections to coexist without deadlock. Note that this number should probably be configured or default to a reasonable number depending on hardware(?)
2. src/include/client.hpp: Added a mutex request_mutex to RpcClient base class, locked in MakeRequest via lock_guard. This protects the Send+Receive sequence from concurrent access, preventing corruption of write_stream, https_result, and other shared state if DuckDB's internal threads ever access the same client concurrently.


hannes:
thanks!

PR #12, opened 2026-04-02
Feature/pushdown support

guillesd:
## Filter & Projection Pushdown for ATTACH Scans

When using `ATTACH` with the RPC extension, queries like `SELECT name FROM remote.test WHERE i > 5` did not pushdown projections or filters to the server.

### Solution

Build SQL on the client side instead of always sending `FROM tablename`. The server already accepts arbitrary SQL via `PREPARE_REQUEST`, so no protocol changes were needed.

For `SELECT name FROM rpc.main.test_data WHERE i > 100`, the client now sends:
```sql
SELECT name FROM test_data WHERE i>100
```
instead of:
```sql
FROM test_data
```

### Implementation

1. **`RpcBindData`** now carries column names/types from the catalog, so the scan phase knows column names to build SQL
2. **`BuildPushdownQuery()`** constructs the SQL at scan init time using DuckDB's `TableFunctionInitInput`:
   - **Projection**: `input.projection_ids` → `SELECT col1, col3` (only output columns, not filter-only columns)
   - **Filters**: `input.filters` → `WHERE col1>100 AND col2='foo'` — uses DuckDB's built-in `TableFilter::ToString()` which already produces valid SQL for constant comparisons, IS NULL/NOT NULL, IN lists, and AND/OR conjunctions
3. **`CanPushdownFilter()`** checks filter types before converting — unsupported types (bloom filters, dynamic filters, etc.) are left for DuckDB to handle locally

### rpc_call and the projection mismatch

Both `rpc_call('uri', 'SQL')` and the ATTACH catalog path share the same `TableFunction` (same pattern as `duckdb-postgres` where `postgres_query` reuses `postgres_scan`'s init/scan functions). Pushdown flags are enabled on both because the catalog path needs them.

This creates a subtlety: for `SELECT a FROM rpc_call('uri', 'SELECT a, b FROM t')`, DuckDB sees `projection_pushdown = true` and creates a 1-column output chunk expecting just `a`. But the server returns 2 columns (`a, b`) because `rpc_call` sends the user's SQL verbatim.

`RpcScan` handles this by checking if the response has more columns than the output expects, and referencing only the needed columns. For the ATTACH path this never triggers because `BuildPushdownQuery` already sends the right projection in the SQL.

### Supported filter types

| Filter | Example SQL |
|--------|------------|
| Constant comparison | `col >= 100`, `col = 'foo'` |
| IS NULL / IS NOT NULL | `col IS NULL` |
| IN list | `col IN (1, 2, 3)` |
| AND / OR conjunctions | `col > 5 AND col < 10` |

### Tests

New `test/sql/pushdown.test` verifies pushdown via `EXPLAIN` plan assertions:
- `Projections: i` appears in the scan node (projection pushdown)
- `Filters: i=42` appears in the scan node (filter pushdown)
- Correctness checks for filtered/projected results

### DIFF
--- a/src/catalog.cpp
+++ b/src/catalog.cpp
@@ -181,6 +181,10 @@ TableFunction RpcTableCatalogEntry::GetScanFunction(ClientContext &context, uniq
 	bind_data->uri = rpc_catalog.GetServerString();
 	bind_data->connection_id = rpc_catalog.GetConnectionId();
 	bind_data->table_name = name;
+	for (auto &col : GetColumns().Physical()) {
+		bind_data->column_names.push_back(col.Name());
+		bind_data->column_types.push_back(col.Type());
+	}
 	bind_data_p = std::move(bind_data);
 	return RpcScanFunction::GetFunction();
 }
--- a/src/include/rpc_bind_data.hpp
+++ b/src/include/rpc_bind_data.hpp
@@ -15,5 +15,7 @@ struct RpcBindData : FunctionData {
 	string table_name;
 	optional_idx estimated_cardinality;
 	unique_ptr<RpcClient> initial_client;
+	vector<string> column_names;
+	vector<LogicalType> column_types;
 };
 } // namespace duckdb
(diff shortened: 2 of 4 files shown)


hannes:
thanks!

Issue #3, opened 2026-03-24
Too much certificate opening during when opening a new TLS connection


guillesd:
I was wondering a good title for this one. While doing my concurrency tests over WebSocket, I run into a "too many open files crash" because every connection was opening the certificate files. I had to stagger connections to make this work but that's not ideal.

## Root (I think)
OnTlsInit [`src/server.cpp` lines 46–72 (OnTlsInit)](https://github.com/guillesd/duckdb-rpc/blob/main/src/server.cpp#L46-L72) loading files everytime. Industry standard seems to be to load the certs on server startup and then reuse the SSL context across connections.

hannes:
Good catch!

hannes:
should be fixed

Issue #2, opened 2026-03-24
WebSocket is currently single threaded


guillesd:
I was running some benchmarks against duckdb-rpc and noticed that the WebSocket implementation handles requests within a single thread, scaling read latencies and total wall time almost linearly.

| Concurrent Readers | Avg Read Latency | p50 | p95 | Wall Time | Total ops/s |
|---|---|---|---|---|---|
| 1 | 8.7ms | 8.5ms | 9.6ms | 0.46s | 109 |
| 4 | 24.5ms | 25.8ms | 28.1ms | 1.38s | 145 |
| 8 | 50.5ms | 51.8ms | 72.1ms | 2.87s | 139 |
| 16 | 130.1ms | 129.8ms | 163.0ms | 26.72s* | 180 |

This doesn't happen with Unix Sockets.

## Root cause

Seems like all connections share one websocketpp event loop thread. `OnMessage` calls `HandleMessage()` synchronously, while one query executes, every other client is blocked:

[`src/server.cpp` lines 121–125 (listen thread)](https://github.com/duckdb/duckdb-rpc/blob/main/src/server.cpp#L121-L125)

```cpp
void WebSocketRpcServer::WebsocketListenThread(WebSocketRpcServer *rpc_server) {
    rpc_server->websocket_server.start_accept();
    rpc_server->websocket_server.run();  // single event loop
}
```

[`src/server.cpp` lines 373–388 (message handler)](https://github.com/duckdb/duckdb-rpc/blob/main/src/server.cpp#L373-L388)

```cpp
void WebSocketRpcServer::OnMessage(const websocketpp::connection_hdl &hdl, const message_ptr &msg) {
    MemoryStream read_stream((data_ptr_t)msg->get_payload().data(), msg.get()->get_payload().size());
    auto received_message = ProtocolMessage::FromMemoryStream(read_stream);
    auto response_message = HandleMessage(*received_message);  // blocks the entire event loop

    MemoryStream write_stream;
    response_message->ToMemoryStream(write_stream);
    websocket_server.send(hdl, write_stream.GetData(), write_stream.GetPosition(),
                          websocketpp::frame::opcode::binary);
}
```

hannes:
OnMessage should probably create a worker thread. Hang on :)

guillesd:
I've tested the latest build and it does seem to work!

PR #17, opened 2026-04-14
Some improvement on batching logic to reduce the amount of fetch requests

guillesd:
This PR modifies slightly the logic to be able to fetch a vector of data chunks. The heuristics are very simple and depend on two exposed settings:
- `quack_fetch_batch_chunks`: max chunks per batch. Default 12.
- `quack_fetch_batch_bytes`: approximate payload cap in bytes. Default 4 MiB.

The default values are based on a paremeter sweep with `quack` running in localhost. I suspect optimal values may be different if you are running the server remote, and they also may change if you are running both the sever and the client in the same availability zone in the cloud. Further investigation can go into this route. The results of the parameter sweep (using `FROM lineitem` query with sf=10):

```
  ┌─────────────────────────────┬───────────┬────────────────────────────────────┬─────────────────────────────┐
  │           Config            │ Wall time │            FETCH calls             │         vs baseline         │
  ├─────────────────────────────┼───────────┼────────────────────────────────────┼─────────────────────────────┤
  │ Baseline (1 chunk/FETCH)    │ 4.95 s    │ ~29,290                            │ —                           │
  ├─────────────────────────────┼───────────┼────────────────────────────────────┼─────────────────────────────┤
  │ 4 chunks × 2 MiB            │ 4.56 s    │ —                                  │ –7.9%                       │
  ├─────────────────────────────┼───────────┼────────────────────────────────────┼─────────────────────────────┤
  │ 8 chunks × 2 MiB            │ 4.42 s    │ —                                  │ –10.7%                      │
  ├─────────────────────────────┼───────────┼────────────────────────────────────┼─────────────────────────────┤
  │ 12 chunks × 4 MiB (default) │ 4.42 s    │ 2,452 (measured)                   │ –10.7% / ~12× fewer FETCHes │
  ├─────────────────────────────┼───────────┼────────────────────────────────────┼─────────────────────────────┤
  │ 16 chunks × 4 MiB           │ 4.52 s    │ —                                  │ –8.7%                       │
  ├─────────────────────────────┼───────────┼────────────────────────────────────┼─────────────────────────────┤
  │ 64 chunks × 16 MiB          │ 5.10 s    │ —                                  │ +3.0% (regression)          │
  └─────────────────────────────┴───────────┴────────────────────────────────────┴─────────────────────────────┘
```

> I did not fetch the amount of requests with the other parameters because I only thought of it after running this benches and I didn't want to run everything again. But the important ones are there.

### Other avenues to explore

- Prefetching on the client: this is only meaningful if the actual fetch requests take some time, so in the localhost checks it improved nothing. The idea is that we can have a sequence like this one `[send FETCH #1] → [send FETCH #2] → [decode #1] → [send FETCH #3] → [decode #2] → ...` where we send a fetch request async before the decoding of the previous request have been done as to not have client threads waiting. This could be interesting because it is low impact in the current code.

- Body compression? DataChunk::Serialize already compresses dictionary / constant / sequence vectors, but flat vectors (the common case for table scans) go out uncompressed AFAIK (I'm not sure).

- The general bottleneck is the mutex in RPCConnection to prevent multiple threads from reading a streaming `QueryResult` in parallel, which is not currently possible (I believe). I think potentially rethinking this will be the biggest gain, since the parallelization at the client side hits this single bottleneck at the server side.

### DIFF
--- a/src/catalog.cpp
+++ b/src/catalog.cpp
@@ -132,10 +132,12 @@ unique_ptr<ColumnDataCollection> RpcCatalog::ExecuteCommand(const string &query)
 	chunk_collection->Initialize(response->Types());
 	while (true) {
 		auto fetch_response = client->Request<FetchResponseMessage>(make_uniq<FetchRequestMessage>(connection_id));
-		if (!fetch_response || !fetch_response->ResponseData() || fetch_response->ResponseData()->size() == 0) {
+		if (!fetch_response || fetch_response->Chunks().empty()) {
 			break;
 		}
-		chunk_collection->Append(*fetch_response->ResponseData());
+		for (auto &chunk : fetch_response->Chunks()) {
+			chunk_collection->Append(*chunk);
+		}
 	}
 	return chunk_collection;
 }
--- a/src/include/message.hpp
+++ b/src/include/message.hpp
@@ -186,17 +186,22 @@ class FetchResponseMessage : public ProtocolMessage {
 public:
 	static constexpr MessageType TYPE = MessageType::FETCH_RESPONSE;
 
-	explicit FetchResponseMessage(unique_ptr<DataChunk> response_data_p)
-	    : ProtocolMessage(TYPE), response_data(std::move(response_data_p)) {};
+	FetchResponseMessage() : ProtocolMessage(TYPE) {};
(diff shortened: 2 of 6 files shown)


hannes:
thanks!

PR #20, opened 2026-04-15
Client: Avoid builtin httplib, go with HTTPUtils from DatabaseInstance

carlopi:
This is similar to (and substitute) https://github.com/duckdb/duckdb-quack/pull/16, modelled after https://github.com/duckdb/duckdb-iceberg/pull/576.

After that, a few other changes for errors I bumped against.
* add httpfs as dependency
* fixup tests to require https (autoload) + adapt them to httpfs logging
* do check (before `rpc_start`) whether the port is already taken by a different process. Same process, the check do not works, but that should be less of an issue.

### DIFF
--- a/extension_config.cmake
+++ b/extension_config.cmake
@@ -9,4 +9,3 @@ duckdb_extension_load(quack
 # Any extra extensions that should be built
 duckdb_extension_load(json)
 duckdb_extension_load(autocomplete)
-duckdb_extension_load(httpfs)
--- a/extension_config.cmake
+++ b/extension_config.cmake
@@ -9,3 +9,4 @@ duckdb_extension_load(quack
 # Any extra extensions that should be built
 duckdb_extension_load(json)
 duckdb_extension_load(autocomplete)
+duckdb_extension_load(httpfs)
--- a/extension_config.cmake
+++ b/extension_config.cmake
@@ -9,3 +9,8 @@ duckdb_extension_load(quack
 # Any extra extensions that should be built
 duckdb_extension_load(json)
 duckdb_extension_load(autocomplete)
+
+duckdb_extension_load(httpfs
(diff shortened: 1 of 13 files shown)


src/client.cpp
```diff
@@ -1,8 +1,37 @@
...
+
+namespace duckdb {
+// Mirror of HTTPFSParams from the httpfs extension (duckdb-httpfs: src/include/httpfs_client.hpp).
+// Redeclared here to avoid a source-level include dependency on httpfs's FetchContent tree.
+// Upstream treats this layout as ABI (duckdb-wasm reimplements it and new fields must be appended
+// at the end). We only touch ca_cert_file / enable_server_cert_verification, but every preceding
+// field must stay in lockstep. Keep in sync with upstream.
+class HTTPState;
+struct RpcExtensionHTTPFSParams : public HTTPParams {
+	RpcExtensionHTTPFSParams(HTTPUtil &http_util) : HTTPParams(http_util) {
```
guillesd: Is this whole thing necessary? There are some parameters here that don't have anything to do with RPC, like `s3_version_id_pining`

extension_config.cmake
```diff
@@ -8,4 +8,9 @@ duckdb_extension_load(quack
 
 # Any extra extensions that should be built
 duckdb_extension_load(json)
-duckdb_extension_load(autocomplete)
\ No newline at end of file
+duckdb_extension_load(autocomplete)
+
+duckdb_extension_load(httpfs
```
guillesd: Humble question: is there a compelling reason why we should generally go through the httpfs extension? Should httpfs be the thing that duckdb uses to interact with any service via http? 

I thought that the idea of httpfs was more built for file systems over the network (or things like s3). 

Mi initial idea was to extend httputil from core.. but you probably know better why that is not a better idea

PR #31, opened 2026-04-20
Add catalog enumeration in attach path

guillesd:
It all started wanting to make this work:

```sql
ATTACH 'quack:localhost:9090' AS rpc (disable_ssl true);
USE rpc;
SHOW TABLES;
┌───────────┐
│   name    │
│  varchar  │
├───────────┤
│ another_t │
│ t         │
└───────────┘
```

> I go through the regular `PrepareRequestMessage`. We could argue that this could be a different instruction/request type.

### First rabbit hole - transactions

`RpcTransaction::Start()` was being called eagerly every time DuckDB opened a transaction. This conflicted with calling `FROM duckdb_tables()`, I think in the server side duckdb_tables was probably initiating a transaction already? 

New path is lazily initiate transactions, only when they are actually needed. The read path will not really initiate a transaction on the server side. This approach is based on what I've seen in duckdb-postgres.

### Second rabbit hole - caching catalog entries

The idea is that you cache the RPC catalog once when ATTACHing, were we fetch schema names from `information_schema.schemata`, then for each schema we eagerly call `LoadTableCache` — a single query to `duckdb_columns()` that returns all table names + column names + types in one RPC round-trip. Results are stored in a `case_insensitive_map_t<unique_ptr<CatalogEntry>>` per schema, guarded by `table_cache_lock`. Once loaded,  subsequent calls to `LoadTableCache` return immediately.

Then of course there is the problem if you have multiple clients attaching. Right now for this cache miss reload, if `LookupEntry` doesn't find a table after the cache is loaded, it calls `ReloadTableEntry` which fetches that single table's metadata from the server. So if another client created a table, you can query it by name and it'll be discovered on the fly. This is okeyish, but still doesn't solve the problem that if `client a attached > CREATE TABLE a` then `client b attached > CREATE TABLE b` then `client a > SHOW TABLES` shows only table `a`.

A couple of ways to solve this:
- Refetch on every query. The overhead may be negligible, depends on the amount of tables x columns.
- Have a function that refreshes all cache or clears it, then it will be reloaded again.
- Refresh cache on specific queries (like duckdb_tables/show tables)

> I am afraid of transactions so please review this!

### DIFF
--- a/src/catalog.cpp
+++ b/src/catalog.cpp
@@ -289,34 +289,6 @@ void RpcSchemaCatalogEntry::RemoveFromTableCache(const string &table_name) {
 	table_cache.erase(table_name);
 }
 
-optional_ptr<CatalogEntry> RpcSchemaCatalogEntry::ReloadTableEntry(const string &table_name) {
-	auto &rpc_catalog = catalog.Cast<RpcCatalog>();
-	auto &context = rpc_catalog.GetContext();
-
-	auto query = StringUtil::Format("SELECT column_name, data_type "
-	                                "FROM duckdb_columns() "
-	                                "WHERE schema_name = %s AND table_name = %s AND NOT internal "
-	                                "ORDER BY column_index",
-	                                KeywordHelper::WriteQuoted(name, '\''),
-	                                KeywordHelper::WriteQuoted(table_name, '\''));
-	auto result = rpc_catalog.ExecuteCommand(query);
-	auto rows = result->GetRows();
-
-	if (rows.size() == 0) {
-		return nullptr;
-	}
-
-	CreateTableInfo create_info(*this, table_name);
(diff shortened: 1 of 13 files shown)


guillesd:
Also there seems to be some flakyness in the tests in 1.5 branch which seem resolved in my branch.. not sure why.

guillesd:
@hannes Carlo mentioned that the PR does too much (touches transactions but also this entry loading stuff). Would it be better if I split this? The transaction work does fix some problems like https://github.com/duckdblabs/duckdb-internal/issues/8956

guillesd:
I did also homogenised the tests. Every test now uses this foreach server thing and gets a different port, otherwise when running them together I was getting weird conflicts.

guillesd:
To be picked up in the future

PR #41, opened 2026-04-22
Fix rpc scan when remapping

carlopi:
Fix for issue found in testing, fixes cases like:
```sql
SELECT b FROM rpc_call('quack:localhost', 'SELECT 1 as a, ''2'' as b', disable_ssl=true);
```

On top of https://github.com/duckdb/duckdb-quack/pull/40


hannes:
thanks!

PR #43, opened 2026-04-29
Use c++17 in duckdb and extension

smvv:


smvv:
dropping c++17 solved the enum linking issue.

PR #45, opened 2026-05-01
Authz rewrite

guillesd:
This PR does two main things:
1. Populates the authorization callback to `CATALOG_REQUEST` and `APPEND_REQUEST`.
2. Refactors authorization to return a query, enabling users to provide authorization functions that modify queries. This is cool but could be funky (gun for SQL injection?).

> [!IMPORTANT]
> Very important behavioural change is that that raising an Authorization error is now on the user. They need to throw `error('blabla')` in order for the Authorization to fail. To clarify, this does not happen in the Authentication callback, where the function returns a boolean and the error message is generic and handled outside of the callback. 

There is a small caveat to this PR. Since we don't have `APPEND_REQUEST_BEGIN`/`APPEND_REQUEST_END` in the protocol, `APPEND_REQUEST` runs the authz callback for every message. I added `AppendAuthzCacheEntry` per `RpcConnection` to workaround this issue, but we may consider in the future for `INSERT`, `DELETE`, `UPDATE`, etc. that go through the catalog path, to enable BEGIN/END requests?

## Example usage

A single SQL macro switches on `op_kind` to enforce a read-only policy:

- `query` → block writes/DDL via a regex on the leading keyword.
- `create_table` / `create_schema` → unconditional block.
- `drop_table` → unconditional block.
- `append` → unconditional block.

```sql
LOAD quack;
LOAD httpfs;
SET GLOBAL rpc_default_token = 'demo-token';

CREATE MACRO authz(sid, op, q) AS (
    CASE
        WHEN op = 'query'
             AND regexp_matches(q, '^\s*(insert|update|delete|merge|create|drop|alter|truncate|copy)\b', 'i')
                                                              THEN error('writes blocked')
        WHEN op = 'create_table' OR op = 'create_schema'      THEN error('ddl blocked')
        WHEN op = 'drop_table'                                THEN error('drops disabled')
        WHEN op = 'append'                                    THEN error('writes disabled')
        ELSE q
    END
);
-- Only need GLOBAL if running in the same terminal server and client
SET GLOBAL rpc_authorization_function = 'authz';

CREATE TABLE existing_tbl(i INTEGER);
INSERT INTO existing_tbl VALUES (1), (2);

CALL rpc_start('quack:localhost:31999');
ATTACH 'quack:localhost:31999' AS r;
```

### Reads pass through

```sql
FROM rpc_call('quack:localhost:31999', 'select 42');
```
```
┌───────┐
│  42   │
│ int32 │
├───────┤
│    42 │
└───────┘
```

```sql
FROM r.existing_tbl;
```
```
┌───────┐
│   i   │
│ int32 │
├───────┤
│     1 │
│     2 │
└───────┘
```

### PREPARE rejection — write via `rpc_call`

```sql
FROM rpc_call('quack:localhost:31999', 'insert into existing_tbl values (99)');
```
```
IO Error: Expected PREPARE_RESPONSE message, got error message instead:
Authorization failed: {"exception_type":"Invalid Input","exception_message":"writes blocked"}
```

### CATALOG rejection — `CREATE TABLE` through ATTACH

```sql
CREATE TABLE r.t(i INTEGER);
```
```
IO Error: Expected CATALOG_RESPONSE message, got error message instead:
Authorization failed: {"exception_type":"Invalid Input","exception_message":"ddl blocked"}
```

### CATALOG rejection — `DROP TABLE` through ATTACH

```sql
DROP TABLE r.existing_tbl;
```
```
IO Error: Expected CATALOG_RESPONSE message, got error message instead:
Authorization failed: {"exception_type":"Invalid Input","exception_message":"drops disabled"}
```

### APPEND rejection — `INSERT` through ATTACH

```sql
INSERT INTO r.existing_tbl VALUES (3);
```
```
IO Error: Expected APPEND_RESPONSE message, got error message instead:
Authorization failed: {"exception_type":"Invalid Input","exception_message":"writes disabled"}
```

### DIFF
--- a/src/rpc_server.cpp
+++ b/src/rpc_server.cpp
@@ -70,6 +70,48 @@ static bool EvaluateAuthQuery(DatabaseInstance &db, const string &sql, ARGS... v
 	return true;
 }
 
+// Run the configured authorization callback as `SELECT <fn>(?, ?, ?)` and
+// return the VARCHAR it produced. Throws on any error from the callback (e.g.
+// `error('msg')`), on missing rows, or on a non-VARCHAR/NULL return.
+static string EvaluateAuthorize(DatabaseInstance &db, const string &sid, const string &op_kind,
+                                const string &payload) {
+	Connection dummy_connection(db);
+	auto sql = StringUtil::Format("SELECT %s(?, ?, ?)", GetSettingString(db, "rpc_authorization_function"));
+	auto result = dummy_connection.Query(sql, Value(sid), Value(op_kind), Value(payload));
+	if (!result || result->HasError()) {
+		throw InvalidInputException(result ? result->GetError() : "authorization callback failed");
+	}
+	auto chunk = result->Fetch();
+	if (!chunk || chunk->size() == 0) {
+		throw InvalidInputException("authorization callback returned no rows");
+	}
+	auto val = chunk->GetValue(0, 0);
+	if (val.IsNull() || val.type().id() != LogicalTypeId::VARCHAR) {
+		throw InvalidInputException("authorization callback must return VARCHAR");
+	}
+	return val.GetValue<string>();
+}
+
+// Cached APPEND authz: fires the callback once per (connection, schema.table),
+// this is to avoid repeated calls for every chunk in a large append.
+static const AppendAuthzCacheEntry &AuthorizeAppendCached(DatabaseInstance &db, RpcConnection &conn,
+                                                          const string &sid, const string &target) {
+	auto it = conn.append_authz_cache.find(target);
+	if (it != conn.append_authz_cache.end()) {
+		return it->second;
+	}
+	AppendAuthzCacheEntry entry;
+	try {
+		EvaluateAuthorize(db, sid, "append", target);
+		entry.allowed = true;
+	} catch (const std::exception &e) {
+		entry.allowed = false;
+		entry.reject_message = string("Authorization failed: ") + e.what();
+	}
+	auto inserted = conn.append_authz_cache.emplace(target, std::move(entry));
+	return inserted.first->second;
+}
+
 string RpcServer::GenerateSessionId() {
 	return StringUtil::GenerateRandomName(32);
 }
@@ -175,10 +217,12 @@ unique_ptr<ProtocolMessage> RpcServer::HandleMessageInternal(ProtocolMessage &re
 			return make_uniq<ErrorMessage>("Invalid connection id");
 		}
 
-		if (!EvaluateAuthQuery(
-		        *db, StringUtil::Format("SELECT %s(?, ?)", GetSettingString(*db, "rpc_authorization_function")),
-		        Value(prepare_request_message.ConnectionId()), Value(prepare_request_message.Query()))) {
-			return make_uniq<ErrorMessage>("Authorization failed");
+		string query_to_run;
+		try {
+			query_to_run = EvaluateAuthorize(*db, prepare_request_message.ConnectionId(), "query",
+			                                 prepare_request_message.Query());
+		} catch (const std::exception &e) {
+			return make_uniq<ErrorMessage>(string("Authorization failed: ") + e.what());
 		}
 
 		std::unique_lock<std::mutex> lock(rpc_connection->lock);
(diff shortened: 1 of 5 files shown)


src/rpc_server.cpp
```diff
@@ -70,6 +70,48 @@ static bool EvaluateAuthQuery(DatabaseInstance &db, const string &sql, ARGS... v
...
+}
+
+// Cached APPEND authz: fires the callback once per (connection, schema.table),
+// this is to avoid repeated calls for every chunk in a large append.
+static const AppendAuthzCacheEntry &AuthorizeAppendCached(DatabaseInstance &db, RpcConnection &conn,
+                                                          const string &sid, const string &target) {
+	auto it = conn.append_authz_cache.find(target);
+	if (it != conn.append_authz_cache.end()) {
+		return it->second;
+	}
```
carlopi, line 105: I would at the very least make `sid` part of the map. This otherwise allows attackers that snooped the connection_id to just append data.

Also, having `target` only I am not sure, does it mean that a next query (still appending) also gets auth for free?

carlopi: Actually, now I understood this is a second line, token is still evaluated

src/rpc_server.cpp
```diff
@@ -70,6 +70,48 @@ static bool EvaluateAuthQuery(DatabaseInstance &db, const string &sql, ARGS... v
 	return true;
 }
 
+// Run the configured authorization callback as `SELECT <fn>(?, ?, ?)` and
+// return the VARCHAR it produced. Throws on any error from the callback (e.g.
+// `error('msg')`), on missing rows, or on a non-VARCHAR/NULL return.
+static string EvaluateAuthorize(DatabaseInstance &db, const string &sid, const string &op_kind,
+                                const string &payload) {
+	Connection dummy_connection(db);
+	auto sql = StringUtil::Format("SELECT %s(?, ?, ?)", GetSettingString(db, "rpc_authorization_function"));
```
carlopi, line 82: I think there are safer way that avoid injection.

guillesd: Tell me how plz!

src/rpc_server.cpp
guillesd: Yes token is still evaluated

src/rpc_server.cpp
carlopi: I think the authorization code does prepared statement + execute prepared statement? That has to work for sure

guillesd: EvaluateAuthQuery and EvaluateAuthorize do the same actually, the difference is that the `StringUtil` part in EvaluateAuthQuery is already done when calling the function, which is confusing. I will update it!

carlopi:
I did some more prototyping around this, maybe would it make sense to have:
```
authn(varchar payload, map(varchar,varchar) request_info, map(varchar, varchar) server_info) -> returns struct (varchar, map(varchar, varchar))
authz(varchar payload, map(varchar,varchar) request_info, map(varchar, varchar) server_info, map(varchar,varchar) authn_info) -> returns struct (varchar, map(varchar, varchar))
```

Idea is that instead of adding values one by one when needed, we bundle them in a `map`, and then signature is flexible over evolutions.

`server_info` can be populated by whatever is relevant and known server side (current timestamp, scope, sid, etc), `request_info` is populated directly from an optional on the wire field (whatever the client decides to send it's added there, for example timestamp when request was generated, possibly some integrity check like sha, etc). `payload` for authn is the bearer token, `payload` for authz is like the SQL in case of query or whatever string representation in other cases. Return value is a struct with a boolean saying pass | fail, and a map with extra info. The map generated on success by `authn` is passed down to `authz` as authz_info.

I think this is the most flexible, and does allow for example to implement sigv4 like signatures, and allow this to evolve over time while staying flexible

guillesd:
@carlopi sounds good! If you want to give it a crack, then the only thing that should be at least reused from this PR is to populate the callback in different places. Currently it was only firing on `CONNECTION_REQUEST` (and maybe some other) but not on all other protocol messages.

Also the previous auth function was a `template` I think precisely because of this. If you think the `map` option is better, go for it.

PR #46, opened 2026-05-01
not installing httpfs on every request

hannes:


src/quack_extension.cpp
```diff
@@ -156,6 +157,12 @@ static void LoadInternal(ExtensionLoader &loader) {
 	loader.RegisterFunction(RpcStartFunction::GetFunction());
 	loader.RegisterFunction(RpcStopFunction::GetFunction());
 
+	auto& db = loader.GetDatabaseInstance();
+	ExtensionHelper::AutoLoadExtension(db, "httpfs");
+	if (!db.ExtensionIsLoaded("httpfs")) {
+		throw MissingExtensionException("The rpc extension requires the httpfs extension to be loaded!");
+	}
+
```
carlopi, line 165: Autoload during extension load is particularly crap, and I would avoid that.

One classic workaround: bool is_httpfs_loaded = false, and do the check only once.

PR #47, opened 2026-05-02
Let's /quack, change default endpoint from /rpc to /quack

carlopi:
Very minor, in my taste we could as well change the endpoint to quack, so while looking at traffic it's clearer that the two instances quack to each other.


hannes:
aye!

PR #51, opened 2026-05-05
GenerateRandomData via EncryptionUtils

carlopi:
Move `GenerateSessionId` to use `EncryptionUtils`'s provided secure random functionality (possibly autoloading `httpfs` if not there yet).

SQL side, it should be equivalent. Note: I have not benchmarked the impact of random generator with a workload heavy on creating different connections, I am not sure if that might be visible, but I would say while checking how to improve (possibly, bulk generating random bytes and consuming caching them) it would make sense, secure has to be the default here.


hannes:
thanks

PR #53, opened 2026-05-05
auth: separate cached Connection per callback

carlopi:
EvaluateAuthn / EvaluateAuthz each own a Connection + cached PreparedStatement instead of allocating a fresh Connection per call. Caches are independent. On execution-level error the cache is reset so the next call rebuilds from scratch.

On a random benchmark, that is figuring out what's the latency of doing `FROM rcp_call('quack:localhost', 'SELECT 42');`, I got a significant (-30% speedup) via avoiding re-creating the connection AND preparing the statement in advance.

Forces SetScope::GLOBAL on rpc_authentication_function and rpc_authorization_function — the server's eval connection has its own ClientContext, so a session-local SET would silently no-op for the auth path. GLOBAL ensures changes actually take effect.

Note, this does creates some possible issues when authn / authz callbacks if they were misbehaving, like if they have side-effects AND if they happen to fail (like, be semantically incorrect) from time to time. It might be worth taking the hit now, since later it might be harder to restrict this.

### DIFF
--- a/src/include/rpc_server.hpp
+++ b/src/include/rpc_server.hpp
@@ -43,6 +43,9 @@ public:
 
 	string GenerateSessionId();
 
+	bool EvaluateAuthn(const Value &v1, const Value &v2);
+	bool EvaluateAuthz(const Value &v1, const Value &v2);
+
 	virtual ~RpcServer();
 
 protected:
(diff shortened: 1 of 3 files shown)


carlopi:
Here main question is if the change is desirable, with the tradeoff being speed vs less clear caching semantics. Maybe it could be done behind a setting (also global)? Like "SET authn_authz_cache_connection = false;` goes back to recreating them on the fly every time? That could allow easier debugging when this might be unpredictable.

hannes:
will this not bite us in highly concurrent scenarios?

guillesd:
Also I guess we are ditching the idea that Authorization callback can modify the query? Or should we keep this idea for future work?

Btw, something that I tried fixing in my PR, and maybe you want to add to this one, the callbacks only fire on `CONNECTION` and `PREPARE`. It should also be populated to other messages in the protocol.

carlopi:
I'll revisit this later, for now it can be closed.

PR #54, opened 2026-05-05
Add BATCHED_REQUEST | _RESPONSE, that carry multiple messages in one go

carlopi:
More or less trivial, but for the fact that CONNECTION returned sid are saved and then propagated to next ones (if they happen to miss a sid). That should make that patterns like "CONNECTION A, CONNECTION B" + "REQUEST A; REQUEST B" are valid, but also: "CONNECTION A; REQUEST ''", where the REQUEST get's the implicit sid of A

Moves queries like `CONNECTION + REQUEST "SELECT 42"` to perform a single roundtrip, that could potentially mean a 30% speedup on repeated `CALL rpc_call('quack:localhost', 'SELECT 42');`

Main cost: this is very much NOT tested, and there are arbitrary path one could build, and some of them likely make little to no sense.

### DIFF
--- a/src/include/message.hpp
+++ b/src/include/message.hpp
@@ -19,6 +19,8 @@ enum class MessageType : uint8_t {
 	CATALOG_RESPONSE = 10,
 	APPEND_REQUEST = 11,
 	APPEND_RESPONSE = 12,
+	BATCH_REQUEST = 13,
+	BATCH_RESPONSE = 14,
 	ERROR = 100
 };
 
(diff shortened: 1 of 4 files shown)


carlopi:
This is a draft since has quite some surface and little validation. Maybe let's figure out if this is interesting to land or not first.

carlopi:
This fails a test that uses logging:
```
[12/13] (92%): test/sql/logging.test
1. test/sql/logging.test:37
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
================================================================================
unittest is a Catch v2.13.7 host application.
Wrong result in query! (test/sql/logging.test:37)!
Run with -? for options
================================================================================

select count(*) > 0 from duckdb_logs_parsed('RPC') where message_type = 'PREPARE_REQUEST' and query = 'select 42 as answer';
-------------------------------------------------------------------------------
================================================================================
Mismatch on row 1, column (count_star() > 0)(index 1)
0 <> true
================================================================================
Expected result:
================================================================================
true
```
Test needs adapting if this lands.

hannes:
I would not go there yet indeed

carlopi:
Let's close this then. I think worth revisiting at a later point.

PR #56, opened 2026-05-06
More Refactor

hannes:
- remove `rpc_default_token`, we now have a `token` parameter for `quack_serve` and `quack_query` as well as for `attach`. Client-side, the recommended way stays the secret of course, the explicit parameter overrides it though.
- switch to generated serialization/deserialization
- catalog ops now use sql too to sidestep horrid auth problems
- appends can now be checked by the authorization function


hannes:
Also adds #55

PR #57, opened 2026-05-07
Several minor debug fixes

Mytherin:
* Remove broken assertions
* Make `QuackUriParser` function only work on constants, and also emit constants
* Implement `Copy` and `Equals` for `QuackScanBindData`
* Remove `CastNoConst` from `QuackScanBindData`

This PR is accidentally based on `main` so it also pulls in `main`, which is just adding the `README`.


Mytherin:
Not sure why CI isn't running here - maybe our private runners have run out (or it's just another github issue)

PR #59, opened 2026-05-07
Split serialization into `MessageHeader` and `QuackMessage`

Mytherin:
Currently we (de)serialize only a top-level `QuackMessage`. This means that we always deserialize the full message, even if the user is not authenticated. This greatly increases the attack surface of the protocol as unauthenticated users can send messages of *any* message type and force the server to deserialize the message.

This means the potential attack surface is whatever gets deserialized in messages - which could be rather complex as we can deserialize `DataChunks` with arbitrarily nested data in them. It is currently e.g. possible to trigger stack overflows by sending messages with deeply nested data types (`INT[][][][][]...[]`).

Memory exhaustion is also an issue - while we can limit the message size we accept, deserializing a message can increase our memory footprint by several factors more than the message size. Especially if this is in e.g. complex nested types, this all currently sits in non-buffer-managed memory, which means an attacker could balloon memory usage in the server beyond expected amounts and trigger the OOM killer. 

This PR remedies this by limiting what unauthenticated attackers can send to the server. Instead of (de)serializing only a `QuackMessage`, we split up the message serialization into a `MessageHeader` followed by a `QuackMessage`. The header looks as follows:

```cpp
struct MessageHeader {
	MessageType type = MessageType::INVALID;
	string connection_id;
	optional_idx client_query_id;
};
```

If the message type requires authentication (i.e. a valid connection id), we then *first* check if that connection id is valid before proceeding to deserialize the remaining `QuackMessage`. The only message that does not require a connection id is a `ConnectionRequestMessage` - which only contains a single `string` (the authentication string). This means we no longer deserialize complex messages from potentially unauthenticated users which greatly limits the potential attack surface in this area.

From a code perspective, what has changed to make this happen is that we no longer serialize the `QuackMessage` in one go - but rather serialize two independent objects:

```
[MessageHeader]
[QuackMessage]
```

This allows us to deserialize only the header, followed by verifying it is valid, followed by processing the message. From a serialization perspective the main difference here is that the individual messages are independently serialized and deserialized, and we need to manually call `QuackMessage::Deserialize` with the message type as a parameter to figure out which message to deserialize.

### DIFF
--- a/src/include/quack_message.hpp
+++ b/src/include/quack_message.hpp
@@ -44,7 +44,9 @@ private:
 string MessageTypeToString(MessageType type);
 
 struct MessageHeader {
-	MessageHeader(MessageType type_p, string connection_id_p) : type(type_p), connection_id(std::move(connection_id_p)) {}
+	MessageHeader(MessageType type_p, string connection_id_p)
+	    : type(type_p), connection_id(std::move(connection_id_p)) {
+	}
 
 	MessageType type = MessageType::INVALID;
 	string connection_id;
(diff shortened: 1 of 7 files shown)


hannes:
thanks!

PR #60, opened 2026-05-07
Use a timing safe equality for authentication

Mytherin:
This prevents timing attacks from being possible to figure out the token. Ideally we switch to using `CRYPTO_memcmp` from OpenSSL but that would currently incur a dependency. We should add this to `EncryptionUtil` upstream eventually. For now this implementation should be good enough (barring weird compilers that ignore the `volatile` flag and optimize the loop into something else).


hannes:
thanks!

PR #63, opened 2026-05-07
Three minor fixes: normalize URIs, surface failures in bind, handle self stop

carlopi:
3 fixes (last two are connected):
* canonicalize URI, so that `quack:1.2.3.4:9494` and `quack:1.2.3.4` are actually the same
* surface failures in bind
* handle somewhat gracefully self-closing the connection (I am not completely sure the test is deterministic, CI will tell)


Mytherin:
Thanks!

PR #65, opened 2026-05-07
Add `quack_server_list` and make `quack_stop` table-in-out

carlopi:
Combined:
```
FROM quack_server_list() s, quack_stop(s.listen_uri) v;
```
makes for stopping all current active servers.

### DIFF
--- a/src/include/quack_server.hpp
+++ b/src/include/quack_server.hpp
@@ -64,6 +64,15 @@ public:
 		return token;
 	}
 
+	const QuackUri &ListenUri() const {
+		return uri;
+	}
+
+	idx_t ActiveConnectionCount() {
+		std::lock_guard<std::mutex> lock(active_connections_mutex);
+		return active_connections.size();
+	}
+
 	virtual ~QuackServer();
 
 protected:
--- a/src/include/quack_startstop.hpp
+++ b/src/include/quack_startstop.hpp
@@ -14,4 +14,9 @@ public:
 	static TableFunction GetFunction();
 };
 
+class QuackServerListFunction {
+public:
+	static TableFunction GetFunction();
+};
+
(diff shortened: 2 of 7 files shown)


Mytherin:
Wouldn't it work to just make `quack_stop` a table in-out function instead of introducing a new one?

carlopi:
Could a function take both a `TABLE` and a `VARCHAR` overload? I had hit an `INTERNAL Error: Function "quack_stop" has a TABLE parameter, and multiple function overloads - this is not supported`.

Or you are suggesting making `quack_stop` only table-in-out?

Mytherin:
Yeah just making it table in-out only, that should work no?

carlopi:
Only issue I can think it's not nice is that `CALL quack_stop((SELECT 'quack:localhost' as listen_uri))` is harder to write/remember, and it's not symmetric with `CALL quack_serve(...)`.

But I can also see how avoiding x2 is nicer.

carlopi:
Here open question is:
```
--- single entry point, table in-table out
CALL quack_stop((SELECT 'quack:localhost' as listen_uri));
CALL quack_stop((FROM quack_server_list()));
```
or
```
--- two separate entry points, varchar + table in-table out
CALL quack_stop('quack:localhost');
CALL quack_stop_each((FROM quack_server_list()));
```

PR can be adapted for both paths

Mytherin:
Don't table in-out functions also work with scalar parameters? e.g. `range` is a table in-out function

carlopi:
`range` is the right model, I'll try harder to make it work

carlopi:
Thanks @Mytherin for the push, I think it should now be OK.

carlopi:
I'd say this can be though more properly

PR #69, opened 2026-05-07
authn & authz are evaluated on separate connections, so arguably GLOBAL

carlopi:
Also adding a test that might serve as documentation on how to set authn custom functionality via SQL.

@guillesd, would you mind taking a look?

### DIFF
--- a/src/quack_extension.cpp
+++ b/src/quack_extension.cpp
@@ -144,9 +144,9 @@ static void LoadInternal(ExtensionLoader &loader) {
 
 	auto &config = DBConfig::GetConfig(loader.GetDatabaseInstance());
 	config.AddExtensionOption("quack_authentication_function", "Name of a callback function for authentication",
-	                          LogicalType::VARCHAR, Value("quack_check_token"));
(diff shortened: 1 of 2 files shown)


Mytherin:
Thanks!

guillesd:
I'm updating the docs! Thanks!

guillesd:
btw I found that settings first, thens serve doesn't work.

```sql
SET GLOBAL quack_authentication_function = 'check_token';
SET GLOBAL quack_authorization_function  = 'read_only';

CALL quack_serve('quack:localhost', token => 'analytics-team-token');
```

The other way around does:
```sql
CALL quack_serve('quack:localhost', token => 'analytics-team-token');

SET GLOBAL quack_authentication_function = 'check_token';
SET GLOBAL quack_authorization_function  = 'read_only';
```

AI explanation -> quack_serve must come before the auth-function swap, otherwise the server starts with an empty token and fails the new min-length check.

carlopi:
Thanks for checking @guillesd! I'll fix it up.

carlopi:
Fix at https://github.com/duckdb/duckdb-quack/pull/75

PR #75, opened 2026-05-07
auth_token is always emitted on server, independently of authn setting

carlopi:
Based on test provided by @guillesd, basically:
```
SET GLOBAL quack_authentication_function = 'check_token';
SET GLOBAL quack_authorization_function  = 'read_only';

CALL quack_serve('quack:localhost', token => 'analytics-team-token');
```
would not save the explicitly provided `token`.


Mytherin:
Thanks!

PR #76, opened 2026-05-07
Replace allow_other_hostname for allow_unsafe_hostname

carlopi:
Very minor, but I found `allow_other_hostname` a bit off (other than what?), unsure if this is better.


hannes:
its not necessarily unsafe, e.g. in a vpc

carlopi:
Maybe `allow_arbitrary_hostname` or `allow_any_hostname` then? This it's minor, but `other` feels weird (to me).

PR #77, opened 2026-05-07
Avoid issueing Queries if ClientContext is not valid anymore

carlopi:
Note: this silently shallows the queries, but there are no other available course of action I think.

I have seen failures in ROLLBACK, I am not sure if this is now impossible, but I think we can as well add the check.

One question: on the state machine TRANSACTION_STARTED/TRANSACTION_NOT_YET_STARTED/TRANSACTION_FINISHED: what happens if a ROLLBACK or COMMIT are issued SQL side? That I don't currently see.

### DIFF
--- a/src/storage/quack_transaction.cpp
+++ b/src/storage/quack_transaction.cpp
@@ -1,6 +1,7 @@
 #include "storage/quack_transaction.hpp"
 
 #include "duckdb/common/printer.hpp"
+#include "duckdb/main/client_context.hpp"
 #include "storage/quack_catalog.hpp"
 
 namespace duckdb {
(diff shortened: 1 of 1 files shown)


Mytherin:
Thanks!

> TRANSACTION_STARTED/TRANSACTION_NOT_YET_STARTED/TRANSACTION_FINISHED: what happens if a ROLLBACK or COMMIT are issued SQL side? That I don't currently see.

That will run a `COMMIT` or `ROLLBACK` unless no transaction was started yet. If no transaction was started yet we send nothing. This prevents executing "empty" transactions (`BEGIN` -> `COMMIT`).

carlopi:
Ok, what I meant, what if the user SQL is "ROLLBACK" or "COMMIT", are QuackTransactions aware of that? Like, I think we will re-issue "COMMIT" / "ROLLBACK" on our own?

Mytherin:
You mean if they send `COMMIT` or `ROLLBACK` through the `query` method? No, we are not aware of that. I think the answer there is that you shouldn't mix the models - either you use only `query` or you use only the catalog API.

PR #78, opened 2026-05-07
Wrap listen_after_bind in try catch, to cover for failures in httplib's ThreadPool constructor

carlopi:
This might happen if close to machine thread limit, httplib's ThreadPool constructor will not properly handle that. Catch and just fail.


Mytherin:
Thanks!

PR #84, opened 2026-05-08
Add quack_server_list table function (list locally available servers)

carlopi:
Simpler version than https://github.com/duckdb/duckdb-quack/pull/65, no table-in-out machinery.


Mytherin:
Thanks!

PR #96, opened 2026-05-08
Add serialization negotiation

carlopi:
Idea is changing from a fixed serialization version (hardcoded in the extension, and potentially different between server and client) to negotiating the minimum. 

On the 1st request  (`CONNECT_REQUEST`, sent by the client): serialization version is fixed to 7, and as additional payload the max_supported_client (say 7 for `v1.5.2`, but 8 for current `main`) is also sent
On the 1st response (`CONNECTION_RESPONSE`, sent by the server): the negotiaded one (that is min(max_supported_client, max_supported_server)), is used to serialize the message, that does contain explicitly the negotiated version
Following requests (`*_REQUEST | *_RESPONSES`): the negotiated one is used, that should then be compatible.

Note, even on repeated contact to the same server, negotiation happens on every CONNECT, and 7 is used. As an optimization, one can opt-in to a higher version caching by endpoint, but that's not done.

### DIFF
--- a/src/include/quack_client.hpp
+++ b/src/include/quack_client.hpp
@@ -19,6 +19,11 @@ public:
 
 	template <class TARGET>
 	unique_ptr<TARGET> Request(optional_ptr<ClientContext> context, unique_ptr<QuackMessage> request_message) {
+		// Stamp the per-connection negotiated version on the outgoing message. The
+		// CONNECTION_REQUEST branch in ToMemoryStream ignores this and always uses
+		// HANDSHAKE_SERIALIZATION_VERSION, so a freshly-constructed QuackClient with
+		// the default value still produces a decodable handshake.
+		request_message->SetOutgoingSerializationVersion(negotiated_serialization_version);
 		auto response_message = RequestInternal(context, std::move(request_message));
 		if (response_message->Type() != TARGET::TYPE) {
 			if (response_message->Type() == MessageType::ERROR_RESPONSE) {
@@ -31,6 +36,10 @@ public:
 		return unique_ptr_cast<QuackMessage, TARGET>(std::move(response_message));
 	}
 
+	void SetNegotiatedSerializationVersion(idx_t version) {
+		negotiated_serialization_version = version;
+	}
+
 	static unique_ptr<QuackClient> GetClient(DatabaseInstance &db, const QuackUri &uri);
 	static unique_ptr<QuackClient> GetClient(ClientContext &context, const QuackUri &uri);
 
@@ -41,6 +50,7 @@ protected:
 	MemoryStream read_stream, write_stream;
 	DatabaseInstance &db;
 	QuackUri uri;
+	idx_t negotiated_serialization_version = QuackMessage::HANDSHAKE_SERIALIZATION_VERSION;
 
 private:
 	virtual unique_ptr<QuackMessage> RequestInternal(optional_ptr<ClientContext> context,
@@ -50,6 +60,7 @@ private:
 class QuackClientConnection : public enable_shared_from_this<QuackClientConnection> {
 public:
 	explicit QuackClientConnection(unique_ptr<QuackClient> client_p, QuackUri uri_p, string connection_id_p,
+	                               idx_t negotiated_serialization_version_p,
 	                               idx_t max_connections_cached = 1);
 	~QuackClientConnection();
 
@@ -58,6 +69,9 @@ public:
 	}
 	const QuackUri &ServerURI() const {
 		return uri;
+	}
+	idx_t NegotiatedSerializationVersion() const {
+		return negotiated_serialization_version;
 	}
 
 	//! Get a client (either a cached one, or open a new one if required)
@@ -68,6 +82,7 @@ public:
 private:
 	QuackUri uri;
 	string connection_id;
+	idx_t negotiated_serialization_version;
 	mutable mutex lock;
 	mutable vector<unique_ptr<QuackClient>> cached_clients;
 	idx_t max_connections_cached;
--- a/src/include/quack_message.hpp
+++ b/src/include/quack_message.hpp
@@ -60,8 +60,21 @@ struct MessageHeader {
 
 class QuackMessage {
 public:
+	//! Highest serialization version v1.5.x supports. CONNECTION_REQUEST is always
+	//! encoded at this version so a v1.5.x peer can decode the handshake before any
+	//! version negotiation has happened.
+	static constexpr idx_t HANDSHAKE_SERIALIZATION_VERSION = 7;
+
 	void ToMemoryStream(MemoryStream &write_stream) const;
 	static unique_ptr<QuackMessage> FromMemoryStream(MemoryStream &read_stream);
+
+	//! Per-message override of the serialization version used by ToMemoryStream.
+	void SetOutgoingSerializationVersion(idx_t version) {
+		outgoing_serialization_version = version;
+	}
+	idx_t OutgoingSerializationVersion() const {
+		return outgoing_serialization_version;
+	}
 
 	template <class TARGET>
 	TARGET &Cast() {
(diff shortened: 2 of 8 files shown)


carlopi:
For a test: I built 4 different duckdb / quack combinations:

1. duckdb v1.5.2 + quack current
2. duckdb main + quack current (adapted for `main`, see https://github.com/duckdb/duckdb-quack/pull/97 but revert `cc36510bb5d3be7eed7b6adde123b8f0d7861217`)
3. duckdb v1.5.2 + this PR
4. duckdb main + this PR (adapted for `main`, see https://github.com/duckdb/duckdb-quack/pull/97)

Pairing 1+2 and executing:
```
main > CALL quack_serve('quack:localhost', token='aargh');
v1.5.2 > CALL quack_query('quack:localhost', 'FROM range(1231231) SELECT random()::VARCHAR', token = 'aargh');
```
```
Serialization Error:
Failed to deserialize: field id mismatch, expected: 102, got: 107
```

while pairing 3+4 does:
```
main > CALL quack_serve('quack:localhost', token='works');
v1.5.2 > CALL quack_query('quack:localhost', 'FROM range(1231231) SELECT random()::VARCHAR', token = 'works');
```
that does work as expected.

I'll try also 3+3 and 4+4 plus logging what goes on the wire.

Edit, here some beautiful printf debugging:

4+4
```
[quack server] negotiated serialization version = 8 (client_max=8, server_max=8) for session 3762C86ED81EA0AAA0DF60701E471474
[quack client] negotiated serialization version = 8 for quack:localhost (connection_id=3762C86ED81EA0AAA0DF60701E471474)
```

3+3
```
[quack server] negotiated serialization version = 7 (client_max=7, server_max=7) for session 8D3DD1F3F5C092B79452BFF6C2A75C58
[quack client] negotiated serialization version = 7 for quack:localhost (connection_id=8D3DD1F3F5C092B79452BFF6C2A75C58)
```

3+4
```
[quack server] negotiated serialization version = 7 (client_max=8, server_max=7) for session 2E0BEC7F596D723A9AD7E3E1C16E9D60
[quack client] negotiated serialization version = 7 for quack:localhost (connection_id=2E0BEC7F596D723A9AD7E3E1C16E9D60)
```

4+3
```
[quack server] negotiated serialization version = 7 (client_max=7, server_max=8) for session 67140BB5A9A847F4330BDE2C6DF15111
[quack client] negotiated serialization version = 7 for quack:localhost (connection_id=67140BB5A9A847F4330BDE2C6DF15111)
```

carlopi:
@hannes: I think this might also be worth having. Doesn't change much for `v1.5.x` vs `v1.5.x` (or in general between similar enough versions), but makes it compatible cross serialization improvements.

Mytherin:
Maybe this could just be part of the `quack` version? i.e. a given `quack` version should imply a certain serialization version. I think having two separate versions (`quack` version and serialization version) is just adding complexity without much benefit that I can see.

carlopi:
I am not sure, to me they are orthogonal. One is what fields are part of the protocol / what messages are expected to be sent. The other is what serialization version to use.

Say we release `quack` both for `v1.5-variegata` and `main`, they should agree on what messages are there / protocol (that's the `quack` version), but `main` (when talking to `main`) duckdb is fine including more efficient compression formats.

Given they are sort of orthogonal, I don't see a way to force numbering to respect this (and be easy to interpret).

carlopi:
We could postpone this conversation, and just hardcode SerializationVersion to 7 for quack version = 1. That's also fine by me, but I feel this eventually we need to cross.

Mytherin:
I would anchor a quack version on a serialization version (e.g. `quack v1 = serialization version 7`), set that in the serializer / deserializer, and enforce that. We should then create a mapping in the extension (`quack version -> serialization version`).

Mytherin:
That would then mean that if we want to upgrade the serialization version we would have to upgrade the quack version, even if the messages haven't changed, but I don't think that's necessarily a bad thing. Both client and server will support ranges of quack versions so upgrading is rather uneventful.

Mytherin:
In a way I think upgrading the quack version *without* upgrading the serialization version will also be difficult, because messages are serialized using the serialization framework. We don't have any way of gating fields behind anything *besides* the serialization version at this point.

carlopi:
Maybe we check live?

For quack version 1, we just hardcode to 7 (that implies it can't work on v1.4.x), and we think about this more properly at a later stage?

I guess my main issue is: if at some point we bump serialization to 8, all subsequent work CAN'T work with v1.5.x, and that means we basically have more complex time developing the extension if we were to target two versions.

Mytherin:
> I guess my main issue is: if at some point we bump serialization to 8, all subsequent work CAN'T work with v1.5.x, and that means we basically have more complex time developing the extension if we were to target two versions.

I don't really get that - serialization is not "bumped" - core DuckDB can support serialization from `1..7` at this point. When we bump to 8 it will be able to support from `1..8`.

If we want to serialize with serialization version 8 we would need to introduce quack version 2 which will target serialization version 8. That means that quack version 2 cannot be supported by v1.5.2. That is fine, as the client will still be able to support quack version 1, which will use serialization version 7.

I suppose this would be a problem if we would want to somehow do both at once (support serialization version 8 while also incrementing the quack version for v1.5.2) - but I think that's just not really going to happen for two reasons: (1) development for v1.5.2 will stop once v2.0 hits, so not an issue, and (2) serialization version moves slowly and there is not that much to optimize / change long-term. Serialization version 8 has a few performance improvements because we hadn't optimized this path much before - but I don't really see that being a long-term thing where each consecutive serialization version has more performance improvements.

I just don't really see this being a problem long-term - maybe a short-term hiccup where we *might* use a slightly less efficient serialization for a single release - while introducing additional complexity especially w.r.t. using the serialization version for backwards/forwards compatibility *within* quack messages. My main problem there is that we then need to have *two* serialization versions: *quack* serialization version (for messages) and *storage* serialization version (which is used *within* the `DataChunk`, etc).

carlopi:
I guess my issue is: how this evolves:
* quack 1, serialization 7, now
* quack 2, serialization 7, next week, say we add a message type / change some invariant
* quack 3, serialization, 7, next month, again, we change some invariants, both v1.5.x and v2.x
* quack 4, serialization 8, say July (to prep for v2.0 / see if there are serialization bottleneck to be improved), v2.x only
* quack 5, serialization 8, now, we figure out we could improve the protocol / add yet another message type to handle some new requirement, this can be shipped v2.x only

I think the fact that 4 and 5 needs to be in the same numbering scheme, means (to allow min-max to work ranges needs to be continguous) serialization also needs to be increasing.

That means that once 4 lands, we can't do development for v1.5.x, but being incompatible with v1.5.x means less signals from users / etc, basically slower feedback cycle. While I could see (especially on the run-up to v2.0) we might do work both on serialization (where we might discover bottlenecks) and protocol at the same time.

That said, that is maybe not necessary to decide now, like, we could still do (at some point in the future):
* quack 1, serialization fixed to 7, now
* quack 2, serialization fixed to 7, 1 week
* quack 3, serialization fixed to 7, 1 month
* quack 4, serialization = 7 for the first message, serialization negotiated for others (adds a field / protocol for this)

(this would basically some version of this PR, if we feel there is a need for this, at a later point)

Long term I think there is sort of the same problem, but obviously need for change will decrease, so yes, it will go somewhat away.

Mytherin:
I can see that - but I think for now this is prematurely solving a problem that can be solved when (if) we actually encounter it as it's only really relevant when incrementing the quack version anyway. Note that an alternative would be to select serialization version based on the DuckDB version that is sent by the client / server in that situation (albeit less flexible).

carlopi:
I am OK also to piggyback on DuckDB version (although, that's implicit, I would prefer one of the side to be explicit).

I am also OK to force Serialization (both for Messages but indeed more relevant for DataChunks) to 7, and call it a day. That might lose some perf on `main` vs `main`, but that's better than incompatibility.

carlopi:
I have https://github.com/duckdb/duckdb-quack/pull/99 up, saying Messages (and DataChunks) have to be v1.5.0 compatible.

Mytherin:
Indeed that makes sense - let's go with that for now and address this differently if it needs to be addressed in the future. Thanks.

PR #98, opened 2026-05-09
Restore skipped test, and avoid test/sql/server_list.test in Windows

carlopi:
…(to be investigated)

I am not sure why `test/sql/server_list.test` seems to block somehow in Windows. Needs review, but that should not block publishing.


hannes:
thanks

PR #99, opened 2026-05-10
Force SerializationCompatibility to 7 (compatible with >= v1.5.0)

carlopi:
This makes so `quack` is properly compatible cross duckdb versions, by using the common denominator that is the SerializationCompatibility version for v1.5.0.

Note: this would need to become a function of quack version, but while understanding what's the proper shape, let's keep it simple.


Mytherin:
Thanks!

PR #100, opened 2026-05-10
Fix so that it compiles vs duckdb `main` branch

carlopi:
There are still plenty of deprecation warnings.

This is meant for `main` branch.


Mytherin:
Thanks!

PR #101, opened 2026-05-10
No excluded platforms

carlopi:
Note: I have NOT tested this, it would be handy to do so before merge (possibly pushing on a duckdb-quack branch, and then deleting it).


Mytherin:
Thanks! Started a CI run [here](https://github.com/duckdb/duckdb-quack/actions/runs/25624485119/job/75217000661)

PR #107, opened 2026-05-11
Add setup-vcpkg Make target, and wire in README

carlopi:
Convenience Makefile target, that vendors and setup vcpkg in the extension folder.


Mytherin:
Thanks!

PR #113, opened 2026-05-12
Keep track of query state

ccfelius:
This PR introduces `quacktivity()`, a table function inspired by `pg_stat_activity` to track current active queries.

It shows the `connection_id` from current active connections and their last executed `query`, along with its `state` and `start time` of the query. For now only `active` and `finished` are used in the state. But the idea is that you'll also be able to see when a query is 'idle' or 'cancelled'.

In a follow up PR I want to introduce a function that allows the user to cancel a query based on `connection_id` (like `pg_cancel_backend(pid)`).

Any thoughts and feedback very much appreciated here! I mainly put this PR out for feedback

### DIFF
--- a/src/quack_storage.cpp
+++ b/src/quack_storage.cpp
@@ -55,6 +55,7 @@ vector<QuackConnectionSnapshot> QuackStorageExtensionInfo::GetActiveConnectionSn
 	std::lock_guard<std::mutex> lock(servers_mutex);
 	for (auto &[uri, server] : servers) {
 		for (auto &snapshot : server->GetActiveConnectionSnap()) {
+			snapshot.server_id = uri;
 			result.push_back(std::move(snapshot));
 		}
 	}
--- a/src/include/quack_server.hpp
+++ b/src/include/quack_server.hpp
@@ -40,6 +40,7 @@ struct QuackConnection {
 };
 
 struct QuackConnectionSnapshot {
+	string server_id;
 	string session_id;
 	string sql_query;
 	QuackQueryState query_state = QuackQueryState::IDLE;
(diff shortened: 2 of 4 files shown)


hannes:
Maybe we can boringify the name, quack_activity or something along those lines is more intuitive

ccfelius:
> Maybe we can boringify the name, quack_activity or something along those lines is more intuitive

sounds good!

ccfelius:
Instead of adding this to quack, I think I'll try to implement this in main duckdb instead because it can be useful in general

carlopi:
One thought: it might also have value to prototype here (already for `v1.5.x`), and move it later to duckdb/duckdb. How to make a call if this will end up being easier or harder, unsure.

src/quack_storage.cpp
```diff
@@ -50,6 +50,17 @@ vector<QuackStorageExtensionInfo::ServerSnapshot> QuackStorageExtensionInfo::Lis
 	return result;
 }
 
+vector<QuackConnectionSnapshot> QuackStorageExtensionInfo::GetActiveConnectionSnaps() {
+	vector<QuackConnectionSnapshot> result;
+	std::lock_guard<std::mutex> lock(servers_mutex);
+	for (auto &[uri, server] : servers) {
+		for (auto &snapshot : server->GetActiveConnectionSnap()) {
+			result.push_back(std::move(snapshot));
```
Tishj, line 58: Don't we want to push the server identifier into this, and also expose that in the table function? I can imagine you want to filter the results per server

ccfelius: We can do that for sure yeah!

ccfelius:
> One thought: it might also have value to prototype here (already for `v1.5.x`), and move it later to duckdb/duckdb. How to make a call if this will end up being easier or harder, unsure.

I'm actually thinking to _do_ keep this table function on quack because it might have a slightly different function. I can imagine that in the future, we might want to expose only the active query state from connections that originated from the same client.

PR #127, opened 2026-05-14
Strip quack URI scheme as prefix only

thijs-s:
### Bug

`QuackUri::QuackUri` strips the `quack:` scheme with `StringUtil::Replace`, which replaces **every** occurrence, not just the prefix. So any URI whose host part contains a second `quack:` gets mangled:

| input                       | after `Replace(uri, "quack:", "")` | parsed host | parsed port |
|-----------------------------|------------------------------------|-------------|-------------|
| `quack:localhost:9494`      | `localhost:9494`                   | `localhost` | `9494`      |
| `quack:ilum-quack:9494`     | `ilum-9494`                        | `ilum-9494` | `9494` (default) |
| `quack:quack:9494`          | `:9494`                            | `` (empty)  | `9494`      |

The mangled hostnames don't resolve, so `ATTACH` fails. In practice this breaks the Kubernetes Service-name path for any Service named `<x>-quack`, a natural default for a quack pod and forces a brittle ClusterIP workaround that doesn't survive Service rolling restarts.

### Fix

Replace the two `StringUtil::Replace` calls in `src/quack_uri.cpp` with prefix-only `substr` strips. No behavior change for any URI whose host part does not contain `quack:`.

### DIFF
--- a/src/quack_uri.cpp
+++ b/src/quack_uri.cpp
@@ -8,15 +8,17 @@ QuackUri::QuackUri(string uri_p, bool ssl_p) : ssl(ssl_p), uri(uri_p) {
 	ipv6 = false;
 	port = 9494;
 	StringUtil::Trim(uri);
-	// first off, lets be tolerant and accept this variant, too
+	// strip the scheme as a prefix only — must not use StringUtil::Replace here,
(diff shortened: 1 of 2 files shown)


hannes:
thanks!

PR #128, opened 2026-05-15
`quack_serve()` to return NotImplementedException in wasm

### LINKED ISSUE #125: `quack_serve()` in browser causes internal exception
```
memory D force install quack from core_nightly;
memory D load quack;
memory D call quack_serve();
INTERNAL Error:
Insecure random generation called without setting 'force_mbedtls_unsafe' = true
This error signals an assertion failure within DuckDB. This usually occurs due to unexpected conditions or errors in the program's logic.
For more information, see https://duckdb.org/docs/stable/dev/internal_errors
```

carlopi:
The actual setup for spawning httplib server doesn't work at the moment in Wasm. To avoid weird to user-facing errors and non-sensical situation, just forbid the quack_serve entrypoint.

Fixes https://github.com/duckdb/duckdb-quack/issues/125


Mytherin:
Thanks!

PR #134, opened 2026-05-16
Change default port in URI format examples

yoonghm:
Updated default port number in URI examples from 1294 to 9494.


hannes:
thanks!

PR #137, opened 2026-05-17
Add CancelRequest remote command

### LINKED ISSUE #136: Allow cancellation of remote queries
PREPARE/FETCH on the client side can hang indefinitely on a slow server query — Ctrl-C in the local shell sets the interrupted flag but the scan thread is stuck in a blocking HTTP read.

CurtHagenlocher:
PREPARE/FETCH on the client side could hang indefinitely on a slow server query — Ctrl-C in the local shell set the interrupted flag but the scan thread was stuck in a blocking HTTP read.

Adds MessageType::CANCEL_REQUEST. Server tracks the currently-running client_query_id in an atomic on QuackConnection so the cancel handler can call Connection::Interrupt() without taking the connection lock that the running query is already holding.

On the client, a watchdog thread runs alongside each cancellable Request, polling ClientContext::IsInterrupted() and, on interrupt, opening a fresh client to post the cancel.

Closes #136

### DIFF
--- a/src/quack_client.cpp
+++ b/src/quack_client.cpp
@@ -160,8 +160,7 @@ unique_ptr<QuackMessage> HttpsQuackClient::RequestInternal(optional_ptr<ClientCo
 		// every connection setup / disconnect / append.
 		unique_ptr<InterruptWatchdog> watchdog;
 		if (context && client_query_id.IsValid() && MessageIsCancellable(request_type) && !connection_id.empty()) {
-			watchdog =
-			    make_uniq<InterruptWatchdog>(db, uri, connection_id, *context, client_query_id.GetIndex());
+			watchdog = make_uniq<InterruptWatchdog>(db, uri, connection_id, *context, client_query_id.GetIndex());
 		}
 
 		try {
--- a/src/quack_client.cpp
+++ b/src/quack_client.cpp
@@ -1,3 +1,8 @@
+#include <atomic>
+#include <chrono>
+#include <thread>
+
+#include "duckdb/main/client_context.hpp"
 #include "duckdb/main/database.hpp"
 #include "duckdb/main/extension_helper.hpp"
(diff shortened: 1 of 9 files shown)


src/quack_client.cpp
```diff
@@ -6,6 +11,71 @@
...
 
 namespace duckdb {
+
+namespace {
+
+// Runs alongside a blocking PREPARE/FETCH HTTP call: polls the local ClientContext's
+// interrupted flag and, on observing an interrupt, opens a fresh client and posts a
+// CancelRequest tagged with the same client_query_id. The server matches that id
+// against its active query and calls Connection::Interrupt(), which lets the
+// blocked HTTP call return with an interrupted-query error.
```
CurtHagenlocher, line 21: I couldn't find a documented way to hook into cancellation, so this seemed the most practical alternative for now.

rustyconover:
Hi @CurtHagenlocher this may be relevant https://github.com/duckdb/duckdb-httpfs/pull/324

CurtHagenlocher:
Thanks for the link, @rustyconover. I think this is different in that it's not sufficient just to stop waiting for the HTTP request. I also want to notify the evaluation on the other side of the wire to stop work on the current query. If I understand correctly, this might well be _complementary_ to your change in the sense that canceling the evaluation would ideally also stop any HTTP request _it_ is making, but it would also apply to any longer-running query like a more extreme version of the one in the test case.

And ultimately, this is just following the pattern already established by ODBC, JDBC, ADO.NET, ADBC, etc., of canceling a running query by opening a new connection and issuing a cancel command on it.

rustyconover:
Hi @CurtHagenlocher, yep, I the link I sent is more about cancelling the request the delivery of a cancel command to remote end is also useful (but likely cannot be relied upon).

PR #149, opened 2026-05-22
Bump to DuckDB v1.5.3

carlopi:
That means nightly development will now go out on v1.5.3 core_nightly, reachable via:
```sql
FORCE INSTALL quack FROM core_nightly;
```


Mytherin:
Thanks!

PR #151, opened 2026-05-25
Update README for v1.5.3

szarnyasg:
Rewrite the loading instructions in the text.


carlopi (changes requested):
I think `duckdb` and `extension-ci-tools` should not change.

szarnyasg:
@carlopi, oops, indeed. Thx!

carlopi (approved):
Thanks!

PR #153, opened 2026-05-26
Add `quack_idle_timeout_seconds` options and `quack_status()`

carlopi:
Idea is that there is a new message (still part of the protocol) that's special in the sense that STATUS messages are not updating a global activity timer, and are returning `IDLE` whenever more than `quack_idle_timeout_seconds` idle seconds are elapsed.

Goal is allowing in-protocol polling for status of a given instance (that might return IDLE, signaling it's OK to schedule destroy of the resource).

One thing to note: `quack_idle_timeout_seconds` is a global setting per duckdb-instance, not of a specific server. This is at the moment NOT a per-server property.

What this enable is having a different scheduled service that sends `POST` requests to `<endpoint>/quack`, and interprets the response as either `OK` or `IDLE`. If idle, then resource can be scheduled for destruction.

### DIFF
--- a/src/include/quack_message.hpp
+++ b/src/include/quack_message.hpp
@@ -17,6 +17,8 @@ enum class MessageType : uint8_t {
 	APPEND_REQUEST = 9,
 	SUCCESS_RESPONSE = 10,
 	DISCONNECT_MESSAGE = 11,
+	STATUS_REQUEST = 12,
+	STATUS_RESPONSE = 13,
 	ERROR_RESPONSE = 100
 };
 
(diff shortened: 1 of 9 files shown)


carlopi:
I am not completely sure if there are easier way out, but the in-protocol STATUS_MESSAGE seems the most polished way of getting a signal about a given server global liveliness.

Note that in the same physical machine there might be different duckdb instances, and each duckdb instance might serve 0 or 1 or multiple quack endpoints. `idle` / `ok` is per-instance.

PR #156, opened 2026-05-28
Merge `v1.5-variegata` into main

carlopi:


Mytherin:
This merge should already have been done by https://github.com/duckdb/duckdb-quack/pull/155. The difference between that and this is that this also bumps extension-ci-tools. That doesn't work because `quack` tests can not currently be run in parallel. I've solved that in https://github.com/duckdb/duckdb-quack/pull/159 and also bumped extension-ci-tools, so I think this can be closed now

carlopi:
@Mytherin, I am not sure that's correct, like looking at https://github.com/duckdb/duckdb-quack/blob/main/src/storage/quack_catalog.cpp#L130 in main, it misses a fix from https://github.com/duckdb/duckdb-quack/commit/669440aafd247a892b3827a0945527f274eea45a. I can wait / merge with #159, of feel free to pick it up.

Mytherin:
Ah yeah maybe there's a few more commits on v1.5 now, feel free to rebase

src/quack_activity.cpp
```diff
@@ -0,0 +1,75 @@
...
+		output.SetValue(2, row, snap.sql_query);
+		output.SetValue(3, row, Value(QueryStateToString(snap.query_state)));
+		if (snap.query_state == QuackQueryState::IDLE) {
+			output.SetValue(4, row, Value(LogicalType::TIMESTAMP));
+		} else {
+			output.SetValue(4, row, Value::TIMESTAMP(snap.query_started_at));
+		}
+		row++;
+	}
+	output.SetCardinality(row);
```
Mytherin: `SetChildCardinality` - or this should be rewritten to use `.Append` on the vectors

PR #157, opened 2026-05-28
Keep track of query state

ccfelius:
This PR introduces `quack_activity()`, a table function inspired by pg_stat_activity to track current active queries. This is a primary implementation, and we can add more (detailed) attributes gradually.

`quack_activity()` shows the `connection_id` from current active connections and their last executed query, along with its `state` and `start time` of the query. For now only `active` and `finished` are used in the state. But the idea is that you'll also be able to see when a query is `idle`, `cancelled` or when it an `error` has been thrown.

This is the same PR as https://github.com/duckdb/duckdb-quack/pull/113. But git lost connection between my original fork of quack and the original repo when quack went public.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -13,6 +13,7 @@ add_library(
   quack_start_stop.cpp
   quack_storage.cpp
   quack_uri.cpp
+  quack_activity.cpp
   serialize_quack_message.cpp)
 set(EXTENSION_SOURCES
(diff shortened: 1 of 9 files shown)


Mytherin:
Had a look at this just now after having to work on the test - `quack_activity` is not super descriptive as a name imo, maybe we can rename this to `quack_active_connections` which is more descriptive about what this function returns?

ccfelius:
fine with me

Mytherin:
Cool - I can pick it up

PR #158, opened 2026-05-29
Explicit cancellations

ccfelius:
This PR adds `quack_cancel(server, connection_id, token)`, which explicitly cancels any active query running on a specific connection (`connection_id`). The cancellation is done through a different connection, and will be rejected if the server + token combination or connection_id is incorrect. Cancelling a query by a user interrupt is not covered yet. Since there is already a PR up covering this we might want to integrate https://github.com/duckdb/duckdb-quack/pull/137 too.

**Cancellation speed**
DuckDB cancels when it checks the context flag between different pipeline tasks. Queries like `from sleep_ms(10000)` only consist of a single task and therefore will be cancelled right before they are actually finished. If this is slow for some queries, we should figure out in which cases we should poll the interrupt state more frequently. 

Fixes https://github.com/duckdblabs/duckdb-internal/issues/9423

### DIFF
--- a/src/include/quack_message.hpp
+++ b/src/include/quack_message.hpp
@@ -17,6 +17,8 @@ enum class MessageType : uint8_t {
 	APPEND_REQUEST = 9,
 	SUCCESS_RESPONSE = 10,
 	DISCONNECT_MESSAGE = 11,
+	CANCEL_REQUEST = 12,
+	CANCEL_RESPONSE = 13,
 	ERROR_RESPONSE = 100
 };
 
@@ -381,6 +383,35 @@ protected:
 
 private:
 	ErrorData error;
+};
+
+class CancelRequestMessage : public QuackMessage {
+public:
+	static constexpr MessageType TYPE = MessageType::CANCEL_REQUEST;
+
+	explicit CancelRequestMessage(string connection_id_p, hugeint_t result_uuid_p)
+	    : QuackMessage(TYPE, std::move(connection_id_p)), result_uuid(result_uuid_p) {
+	}
+
+	void Serialize(Serializer &serializer) const override;
+	static unique_ptr<CancelRequestMessage> Deserialize(Deserializer &deserializer);
+
+	hugeint_t result_uuid;
+
+protected:
+	CancelRequestMessage() : QuackMessage(TYPE) {
+	}
+};
+
+class CancelResponseMessage : public QuackMessage {
+public:
+	static constexpr MessageType TYPE = MessageType::CANCEL_RESPONSE;
+
+	explicit CancelResponseMessage() : QuackMessage(TYPE) {
+	}
+
+	void Serialize(Serializer &serializer) const override;
+	static unique_ptr<CancelResponseMessage> Deserialize(Deserializer &deserializer);
 };
 
 } // namespace duckdb
--- a/src/include/quack_message.json
+++ b/src/include/quack_message.json
@@ -65,6 +65,20 @@
     "members": []
   },
   {
+    "class": "CancelRequestMessage",
+    "members": [
+      {
+        "id": 1,
+        "name": "result_uuid",
+        "type": "hugeint_t"
+      }
+    ]
+  },
+  {
+    "class": "CancelResponseMessage",
+    "members": []
+  },
+  {
     "class": "PrepareResponseMessage",
     "members": [
       {
--- a/src/quack_cancel.cpp
+++ b/src/quack_cancel.cpp
@@ -1,0 +1,80 @@
+#include "quack_cancel.hpp"
+#include "duckdb.hpp"
+#include "duckdb/main/database.hpp"
+#include "quack_startstop.hpp"
+#include "quack_storage.hpp"
+#include "quack_client.hpp"
+#include "quack_message.hpp"
+#include "quack_uri.hpp"
+#include "duckdb.hpp"
+
+namespace duckdb {
+
+struct QuackCancelBindData : FunctionData {
+	string target_connection_id;
+	string server_uri;
+	bool finished = false;
+
+	explicit QuackCancelBindData(string connection_id_p) : target_connection_id(std::move(connection_id_p)) {
+	}
+
+	unique_ptr<FunctionData> Copy() const override {
+		auto result = make_uniq<QuackCancelBindData>(target_connection_id);
+		result->finished = finished;
+		return result;
+	}
+	bool Equals(const FunctionData &other_p) const override {
+		return target_connection_id == other_p.Cast<QuackCancelBindData>().target_connection_id;
+	}
+};
+
+static unique_ptr<FunctionData> QuackCancelBind(ClientContext &context, TableFunctionBindInput &input,
+                                                vector<LogicalType> &return_types, vector<string> &names) {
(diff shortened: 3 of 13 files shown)


src/include/quack_message.json
```diff
@@ -64,6 +64,20 @@
     "class": "DisconnectMessage",
     "members": []
   },
+  {
+    "class": "CancelRequestMessage",
+    "members": [
+      {
+        "id": 1,
+        "name": "result_uuid",
```
Mytherin, line 72: A `result_uuid` is only available after a query has (at least partially) been completed. That means that we can't really cancel an in-process query using a `result_uuid`, we can only cancel a result fetch.

A better design here might be to scrap `result_uuid`, and to instead have the *client* provide a `query_id` on the initial `PrepareRequestMessage` message. That way the client already knows the `uuid` of any queries that it has issued and can then cancel them.

That would require a breakage of the `quack` version, however. so that might be better suited for `main` rather than `v1.5-variegata`.

src/include/quack_message.hpp
```diff
@@ -383,4 +385,33 @@ class ErrorResponse : public QuackMessage {
...
+	hugeint_t result_uuid;
+
+protected:
+	CancelRequestMessage() : QuackMessage(TYPE) {
+	}
+};
+
+class CancelResponseMessage : public QuackMessage {
+public:
+	static constexpr MessageType TYPE = MessageType::CANCEL_RESPONSE;
```
Mytherin, line 408: Maybe this can just be a `SuccessResponse`?

src/quack_storage.cpp
```diff
@@ -80,6 +80,18 @@ bool QuackStorageExtensionInfo::StopServer(ClientContext &context, const QuackUr
 	return true;
 }
 
+bool QuackStorageExtensionInfo::CancelConnection(const string &server_id, const string &connection_id) {
```
Mytherin, line 83: Do we need this method?

src/quack_cancel.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+	auto initial_uri = QuackUri(server_uri_str);
+	auto enable_ssl = !initial_uri.IsLocal();
+	if (input.named_parameters.find("disable_ssl") != input.named_parameters.end()) {
+		enable_ssl = !input.named_parameters["disable_ssl"].GetValue<bool>();
+	}
+	auto server_uri = QuackUri(server_uri_str, enable_ssl);
+
+	// Connect to server — validates token via CONNECTION_REQUEST auth function
+	auto client_connection = QuackClient::ConnectToServer(context, server_uri, token);
```
Mytherin, line 53: This function seems to be similar to `quack_query` which does two things: (1) connect to the server, (2) send a cancel request. Maybe we can instead have this behave like `quack_query_by_name` that uses an existing connection instead?

test/sql/quack_cancel.test
```diff
@@ -0,0 +1,48 @@
...
+# group: [sql]
+
+require quack
+
+require httpfs
+
+set ignore_error_messages __none__
+
+statement ok
+CALL quack_serve('quack:localhost', token='asdf');
```
Mytherin, line 12: I think the nicest way of integrating this into `quack` would be for the client to automatically send the server a `CancelRequest` when a query is cancelled client-side (e.g. by the user pressing `Ctrl+C`). That way the server does not continue to do unnecessary processing.

The way this could be achieved is to do this in the destructor of e.g. the `QuackScanBindData`, which could then check if the query completed successfully, and try to send a `CancelRequest` if it hasn't (due to e.g. an exception being thrown).

Mytherin (commented):
Thanks for the PR! Looks great - I've left some comments.

> DuckDB cancels connections lazily; this means that it only checks the context flag between different pipeline tasks. Queries like from sleep_ms(10000) only consist of a single task and therefore will be cancelled right before they are actually finished. In a follow up, we should determine how to improve this, probably by doing something that has been done in https://github.com/duckdb/duckdb-quack/pull/137 such as a 'watchdog' thread that continuously polls the interrupt state.

This is not related to `quack` specifically and shouldn't be solved in `quack` - cancellations are done at the DuckDB level and should be managed in core DuckDB.

The issue here is specifically about the `sleep_ms` function which doesn't check cancellations but just does a large sleep. This could be improved in that specific function if desired by sleeping in a loop instead of doing one large sleep.

src/include/quack_message.json
ccfelius: I agree - for different reasons it would be preferable if the client itself already generates a `query_id` and sends it along directly.

src/quack_cancel.cpp
ccfelius: For this PR I was trying to create a function like `pg_cancel_backend(pid)` (which works very different internally though). 

The idea is that one can cancel a query based on `query_id` (not yet possible) or `connection_id`. This however imposes some security risks since you are able to cancel queries from _a different client_. Curious to hear what your opinion is about this!

In general I think that you don't want a user being able to cancel queries of other users, so my idea was at least to authenticate with the server again. And that is why I explicitly enforced to add the server and token again in `quack_cancel(server, connection_id, token)`.

test/sql/quack_cancel.test
ccfelius: `Ctrl+C` is one way to cancel a query, which also should be addressed. But I was thinking to do that in a separate PR?

src/quack_cancel.cpp
Mytherin: For sure authentication is necessary - what I meant is that functions like `quack_query` (and now `quack_cancel`) do two things: authenticate and do their given action. By using `ATTACH` first the authentication is split off into a separate step and done as part of the `ATTACH`. That simplifies the function implementation since the function doesn't need to do the authentication as well - that split is done in `quack_query_by_name`.

It's not really related to security - both steps are still done - it's just that they're done as part of separate actions (`ATTACH` authenticates, the function runs the command).

test/sql/quack_cancel.test
Mytherin: Sure, that's fine by me

src/quack_cancel.cpp
ccfelius: Yes I understand! But one of the use cases I have in mind is cancelling a query from a different client, in a different duckdb process. In that case, it's not given that you are already authenticated with the server, right?

The client would then first need to attach manually anyways, which means authenticating again. Nevertheless, since the attach syntax is cleaner I will also make it work with `quack_query_by_name`. Maybe I am still missing something here - but you can not use `quack_query_by_name` from a different process right?

Mytherin: It's fundamentally the same - it just means the client needs to do two steps (`ATTACH + quack_query_by_name`) versus just running `quack_query`. You can do both from a different process as well, you just need to run two commands instead of one.

That might sound worse but imo it's better because you're separating concerns. You can get the exact same functionality by running `ATTACH + cancel + DETACH`. If you only implement `quack_query` then you **need** to make a new connection as part of running the function, i.e. there's no possibility of caching / re-using connections, etc. As a result the implementation like `quack_query` is less versatile/more opinionated on how it should be used (one-shot execution only with forced connect / disconnect).

ccfelius: I see! And makes sense. I'll do the `quack_query_by_name` approach then. Thanks for the explanation!

PR #164, opened 2026-06-04
Fix ATTACH for tables with sequence-backed defaults

Jaydeng75:
Related to duckdb/duckdb#22743

## Summary

Quack reconstructs remote table schemas during `ATTACH` by loading the server-side `CREATE TABLE` SQL and rebinding it locally.

For tables containing sequence-backed defaults such as:

```sql
CREATE SEQUENCE id_sequence START 1;

CREATE TABLE test (
    id INTEGER PRIMARY KEY DEFAULT nextval('id_sequence'),
    s VARCHAR
);
```

the local binder attempts to resolve the referenced sequence during catalog reconstruction. Since sequences are not loaded into the attached client catalog, binding fails and `ATTACH` does not complete successfully.

## Solution

Before local binding, detect column defaults that reference `nextval()` and remove those defaults from the client-side schema representation.

Only sequence-backed defaults are stripped; ordinary defaults such as `DEFAULT 42` and `DEFAULT 'hello'` are preserved.

This avoids binding failures caused by server-only sequence objects while maintaining normal schema metadata. Default expressions continue to be evaluated on the server side during inserts.

## Testing

Added regression tests covering:

* ATTACH against tables using `DEFAULT nextval(...)`
* ATTACH against tables using ordinary defaults (`DEFAULT 42`, `DEFAULT 'hello'`)

The sequence-backed default test reproduces the behavior reported in duckdb/duckdb#22743 and verifies that ATTACH succeeds and the remote table can be queried normally.

### DIFF
--- a/src/storage/quack_table.cpp
+++ b/src/storage/quack_table.cpp
@@ -3,6 +3,9 @@
 #include "quack_scan.hpp"
 #include "duckdb/parser/parser.hpp"
 #include "duckdb/parser/statement/create_statement.hpp"
+#include "duckdb/parser/column_list.hpp"
+#include "duckdb/parser/expression/function_expression.hpp"
+#include "duckdb/common/string_util.hpp"
 #include "duckdb/planner/parsed_data/bound_create_table_info.hpp"
 #include "duckdb/parser/parsed_data/create_view_info.hpp"
 #include "storage/quack_view.hpp"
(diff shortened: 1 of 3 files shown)


mrbelderock:
The same happens with timestamp(tz) defaults: DEFAULT now() also leads to a binder error.

Would this fix address such scenarios as well?

Jaydeng75:
Thanks for pointing that out. This PR currently only strips defaults that contain `nextval()`, so it would not address `DEFAULT now()` or other non-sequence defaults that also happen to fail during binding.

I kept the change scoped to the sequence-backed default case reported in duckdb/duckdb#22743. If Quack needs to support additional default expressions that fail client-side binding, I can look into broadening the handling further.

Mytherin:
Thanks for the PR - this has also been fixed in https://github.com/duckdb/duckdb-quack/pull/264, so we don't need this change anymore.

PR #165, opened 2026-06-04
Cancellations by user interrupt and catalog name

ccfelius:
Reworked version of #158.

This PR enables cancellations by `Ctrl+c` on the client, or by `quack_cancel(catalog_name, connection_id)` from a different client. For the latter case, the flow then looks like:

```
ATTACH 'quack:localhost' AS remote (TYPE quack, TOKEN 'asdf');
from quack_cancel('remote', 'some_random_conn_id_data');
```

Also in this PR, the `result_uuid` is scrapped and a `query_uuid` is already generated client side, and sent along with the prepare request to the server.


src/quack_cancel.cpp
```diff
@@ -0,0 +1,75 @@
...
+	data.finished = true;
+	output.SetValue(0, 0, data.target_connection_id);
+	output.SetValue(1, 0, Value::BOOLEAN(data.cancelled));
+	output.SetCardinality(1);
+}
+
+TableFunction QuackCancelFunction::GetFunction() {
+	auto fun =
+	    TableFunction("quack_cancel", {LogicalType::VARCHAR, LogicalType::VARCHAR}, QuackCancelScan, QuackCancelBind);
+	fun.named_parameters["disable_ssl"] = LogicalType::BOOLEAN;
```
Mytherin: nit: this shouldn't be required here anymore

src/quack_server.cpp
```diff
@@ -431,6 +438,16 @@ unique_ptr<QuackMessage> QuackServer::HandleMessageInternal(DatabaseInstance &db
 		}
 		return make_uniq<SuccessResponse>();
 	}
+	case MessageType::CANCEL_REQUEST: {
+		auto &cancel_request_message = received_message.Cast<CancelRequestMessage>();
+		auto &connection = *connection_p;
+		// TODO; we should check the query id in the future
```
Mytherin: It should be possible to do this now, right?

src/quack_scan.cpp
```diff
@@ -47,15 +47,15 @@ static unique_ptr<FunctionData> QuackScanBind(ClientContext &context, TableFunct
...
 	auto bind_response = client.Request<PrepareResponseMessage>(
-	    context, make_uniq<PrepareRequestMessage>(client_connection.ConnectionId(), query));
+	    context, make_uniq<PrepareRequestMessage>(client_connection.ConnectionId(), query, bind_data->query_uuid));
 
 	return_types = bind_response->Types();
 	names = bind_response->Names();
 
 	bind_data->results = std::move(bind_response->MutableResults());
 	bind_data->needs_more_fetch = bind_response->NeedsMoreFetch();
-	bind_data->result_uuid = bind_response->ResultUUID();
```
Mytherin, line 58: I suppose we could check if `bind_response->ResultUUID() == bind_data->query_uuid` for sanity?

Otherwise we might also just remove the `UUID` from `PrepareResponseMessage`

test/sql/cancellations.test
```diff
@@ -0,0 +1,57 @@
...
+# group: [sql]
+
+require quack
+
+require httpfs
+
+set ignore_error_messages __none__
+
+statement ok
+CALL quack_serve('quack:localhost', token='asdf');
```
Mytherin: Can we use the template to start servers so that we don't get failures when running tests in parallel?

Mytherin (commented):
Thanks! Looks great - some minor comments

hannes:
Thanks!

carlopi:
I think this breaks inter-compatibilty between different duckdb & quack versions.

In v1.5.4 + stock quack: `CALL quack_serve('quack:localhost:1116', TOKEN:='12345');`
In main of quack (after this PR): `ATTACH 'quack:localhost:1116' AS d (TOKEN '12345');`

Errors like:
```
IO Error:
Failed to send message: HTTP Error: Request returned HTTP 500 for HTTP POST to 'http://localhost:1116/quack'
```

I am not sure if swapping to ReadPropertyWithDefault or some other strategy allows to keep compatibility, or a bump (maybe behind explicit opt-in by client) is handy.

PR #166, opened 2026-06-04
Get connection id at client

ccfelius:
Adding a scalar function `quack_connection_id('catalog_name')` which returns the current connection id of the attached catalog. Useful for cancellations.


Mytherin:
Thanks!

PR #167, opened 2026-06-05
Fix client_query_id propagation

ccfelius:
This PR fixes the propagation of the `client_query_id`. The `client_query_id` was set _after_ the request message was already serialized, leading to a missing `client_query_id` server side.


Mytherin:
Thanks!

PR #169, opened 2026-06-10
Avoid structured bindings (for now)

carlopi:
Required to avoid Windows CI (in duckdb/duckdb) complaining while bumping quack in v1.5-variegata (see https://github.com/duckdb/duckdb/pull/23178)


Mytherin:
Thanks!

PR #171, opened 2026-06-11
Optimize bulk append

guillesd:
This PR introduces an improvement in bulk append performance by client side batching data chunks before sending them over the network. If the connection is poor (let's say my poor houses 100MBpS) or the host is far away, this can speed up massively (5x in my small benchmark). This does introduce a copy of data chunks into a vector client side, but the threshold caps the memory limit. Even with very wide tables this will never be more than 100MB. We could decide to apply a bytes threshold to the batch, currently it is just based on row count `QUACK_APPEND_FLUSH_ROWS = STANDARD_VECTOR_SIZE * 100ULL;`. This is not exposed as a setting but could be.

> [!IMPORTANT] 
> This PR introduces a protocol change, see below for protocol change discussion.

### Bench client (me in NL) → server (AWS c7g.2xlarge, eu-central-1)
| Workload            | Old (per-chunk) | New (batched) | Speedup |
|---------------------|----------------:|--------------:|--------:|
| SF1 COPY of lineitem (6M rows)  | 643.4 s         | 129.4 s       | 5.0×    |

### Round-trips (measured locally)
This is just simple math, we batch 100 data chunks, so we do 100x less `APPEND_REQUEST`. The larger the table we are copying over, the more considerable this is.

### Follow up comments
I know with some more async/IO stuff dropping in httpfs there are more possibilities for optimization. But I think on the bulk appending side, this is probably the biggest win without Async.

### Regarding protocol versioning

I don't think every protocol change within main (unstable) should get a protocol bump. I just bumped the protocol version to 2 and every subsequent protocol change in main can keep it as is. If you want to run client and server in main, then you can make sure that the builds target the same commit hash. This is what we do in DuckLake with spec versioning. It is not perfect but it allows us to move without having to think about versions in an already unstable environment. Thoughts?


guillesd:
Potentially to make it cleaner, since we use `ColumnDataCollection` on the server side append path, we could introduce:
```c++
//! Wire adapter for a ColumnDataCollection. Serializes the collection CHUNK-BY-CHUNK via the
//! columnar DataChunk::Serialize
class ColumnDataCollectionWrapper {
public:
      explicit ColumnDataCollectionWrapper(unique_ptr<ColumnDataCollection> collection_p)
          : collection(std::move(collection_p)) {
      }

      ColumnDataCollection &Collection() {
              return *collection;
      }

      void Serialize(Serializer &serializer) const;
      static unique_ptr<ColumnDataCollectionWrapper> Deserialize(Deserializer &deserializer);

private:
      unique_ptr<ColumnDataCollection> collection;
};
```
Similar to `DataChunkWrapper`.

guillesd:
I also have a local version with `ParallelSink=true` that speeds this up a bit further (2x vs the current bulk single threaded version), so 10x in total vs the original one. This feels like the same baseline that `FETCH` currently has. This is nice because any follow up improvement coming from the Async IO work will compare against the fastest sync implementation.

Issue #180, opened 2026-06-26
Table created in the rollback-ed transaction is visible


dentiny:
Hi team, I found when I'm using `quack_query_by_name` to execute query, it doesn't explicitly start a transaction at client and server side, so later `ROLLBACK` statement does execute.

Repro SQL statement
```sql
# name: test/sql/quack_query_by_name_rollback_repro.test
# description: quack_query_by_name should participate in the outer transaction rollback
# group: [sql]

require quack

require httpfs

statement ok con1
CALL quack_serve('quack:localhost', token='asdf');

statement ok con2
CREATE SECRET (TYPE quack, TOKEN 'asdf');

statement ok con2
ATTACH 'quack:localhost' AS rpc;

statement ok con2
BEGIN;

statement ok con2
CALL system.main.quack_query_by_name('rpc', 'CREATE TABLE rollback_bug(i INTEGER); INSERT INTO rollback_bug VALUES (42);');

statement ok con2
ROLLBACK;

# Refresh the attached catalog snapshot so we observe the remote server state.
statement ok con2
CALL quack_clear_cache();

# Correct behavior: table creation/insert should have rolled back remotely too.
statement error con2
SELECT * FROM rpc.rollback_bug;
----
Table with name rollback_bug does not exist
```
The error message that I get is
```sh
Query unexpectedly succeeded! (test/sql/quack_query_by_name_rollback_repro.test:32)!
================================================================================
SELECT * FROM rpc.rollback_bug;
================================================================================
i
INTEGER
[ Rows: 1]
42
```

The reason why I have to use `quack_query_by_name` instead of normal SQL statements is [DuckLake does so](https://github.com/duckdb/ducklake/blob/96a4574d65665d15f5ca158892d84493a19d4c62/src/metadata_manager/quack_metadata_manager.cpp#L23-L24).
I file an issue at DuckLake repo as well, along with some analysis: https://github.com/duckdb/ducklake/issues/1267#issuecomment-4806040600

PR #182, opened 2026-06-26
Start transaction before query

dentiny:
Closes https://github.com/duckdb/duckdb-quack/issues/180

Hi team, when I was debugging a DuckLake issue with quack catalog, I found rollback doesn't seem to work properly (see analyze here https://github.com/duckdb/ducklake/issues/1267#issuecomment-4806040600)

In this PR, I explicitly start the transaction at both client and server side, so later rollback does execute.

### DIFF
--- a/src/quack_scan.cpp
+++ b/src/quack_scan.cpp
@@ -89,6 +89,10 @@ static unique_ptr<FunctionData> QuackScanBindCatalogName(ClientContext &context,
 	auto client_wrapper = bind_data->client_connection->GetClient(context);
 	auto &client = client_wrapper->GetClient();
 	bind_data->query_uuid = UUID::GenerateRandomUUID();
+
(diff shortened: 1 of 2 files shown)


dentiny:
The CI failure doesn't seem to be my issue
```sh
/duckdb_build_dir/src/quack_active_connections.cpp:58:32: warning: ‘void duckdb::DataChunk::SetValue(duckdb::idx_t, duckdb::idx_t, const duckdb::Value&)’ is deprecated: Use Vector::Append on data[col_idx] instead (or Vector::SetValue for write-at-index semantics) [-Wdeprecated-declarations]
   58 |                 output.SetValue(0, row, snap.server_id);
      |                 ~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~
```

PR #187, opened 2026-06-28
Add `quack_secret_storage` type (one per AttachedDatabase) and `quack_connection_secret_storage`

carlopi:
Allows:
```sql
ATTACH 'quack:...' AS quack_remote;
CREATE SECRET secret_name IN quack_remote (...);
```
that is implemented as
```sql
CALL quack_query_by_name('quack_remote', 'CREATE TEMPORARY SECRET secret_name IN quack_connection_storage (...)');
```

This adds a `quack_connection_storage` that is tied to a Connection lifetime, this allows automatic management and scoping of secrets tied to Connections, so that for example different client can have a different set of secrets.
Also adds a secret storage named after each quack AttachedDatabase, that stores in the relevant Connection's quack_connection_storage server side the secret after the rewrite.

`quack_connection_storage` can be considered (under a different name) to be upstream into duckdb's `main`.
Creating secret (from a client) into the custom named storage requires server side `quack` to support creating such secrets, so there is some degree of incompatibility, but failures happen clearly on interacting with this feature.

This allows a sever to have multiple clients attached to it, and avoid mix-up in secrets.
Example:
```sql
--- client 1
ATTACH 'quack:...' AS q;
CREATE SECRET s IN q (TYPE http, SCOPE 'https://duckdb.org/', EXTRA_HTTP_HEADERS {'who': 'I am client 1'} );
FROM q.query('FROM read_blob("https://duckdb.org/")');
```
```sql
--- client 2
ATTACH 'quack:...' AS q;
CREATE SECRET s IN q (TYPE http, SCOPE 'https://duckdb.org', EXTRA_HTTP_HEADERS {'who': 'I am client 2'} );
FROM q.query('FROM read_blob("https://duckdb.org/")');
```
have HTTP requests (sent from server `q`) having either "I am client 1" or "I am client 2" as extra headers. Similarly this allows auth or any other secret-based capability to be tied to the actual client.

Note: this DO NOT guarantees any sort of isolation of confidential information, what this allows is avoid overriding or other weird interactions between multiple well-behaved clients that currently require some level of coordination.

Includes also https://github.com/duckdb/duckdb-quack/pull/186

### DIFF
--- a/.github/workflows/MainDistributionPipeline.yml
+++ b/.github/workflows/MainDistributionPipeline.yml
@@ -16,7 +16,7 @@ jobs:
     name: Build extension binaries
     uses: duckdb/extension-ci-tools/.github/workflows/_extension_distribution.yml@v1.5-variegata
     with:
-      duckdb_version: v1.5.3
+      duckdb_version: v1.5.4
       ci_tools_version: v1.5-variegata
       extension_name: quack
 
@@ -24,7 +24,7 @@ jobs:
     name: Code Quality Check
     uses: duckdb/extension-ci-tools/.github/workflows/_extension_code_quality.yml@v1.5-variegata
     with:
-      duckdb_version: v1.5.3
+      duckdb_version: v1.5.4
       ci_tools_version: v1.5-variegata
       extension_name: quack
       format_checks: 'format'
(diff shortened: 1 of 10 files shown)


hannes:
Thanks, I am not convinced that this is the way to go yet. We could also consider a general secret forward in ATTACH.

PR #189, opened 2026-06-29
Server-side scan from client

guillesd:
Today the server applies a remote `INSERT ... SELECT` chunk by chunk via `Connection::Append()`, outside any transaction. Two problems with that: it's not atomic (a statement that fails halfway, e.g. a CHECK violation, leaves the already-applied rows committed), and it re-implements insert logic that core already has.

This PR turns the incoming append stream into a real `INSERT INTO t SELECT * FROM scan_data_from_quack_client('<id>')` that we run on the server, reusing core's `PhysicalBatchInsert` (optimistic row-group writes, memory-bounded buffering, small-batch merging) instead of hand-rolling it. A failing autocommit insert now rolls back to 0 rows for free. (Explicit `BEGIN…ROLLBACK` still needs the client to force-start the transaction; separate follow-up.)

This is server-side only. The current synchronous client is untouched, the async/parallel client comes later.

## How it works

- The client sends each chunk as a `QUACK_SEND_DATA_REQUEST` and ends the stream with a `QUACK_FINALIZE`.
- A small per-stream queue (`QuackDataStream`), keyed by `(connection_id, query_uuid)`, buffers incoming chunks. A `scan_data_from_quack_client` table function is the source of an INSERT we drive on a background thread, popping chunks off that queue. Advertising batch-index partitioning is what makes the planner pick `PhysicalBatchInsert`.
- When the source has no chunk yet, it returns `BLOCKED` + an `AsyncTask` on the ASYNC pool (the same pattern as core's multi-file reader), so the worker thread is parked rather than blocked. It falls back to inline execution when `async_threads=0`.

### Keeping the read/write paths symmetric

I tried to mirror the read path so the two halves converge:
- Same identity: `(connection_id, query_uuid)`, like `PrepareRequest`/`FetchRequest`.
- Same request/response shape: `QUACK_SEND_DATA_REQUEST` answered by `QUACK_SEND_DATA_RESPONSE` (a placeholder today, with a field reserved for a future server "accept budget"). Errors keep coming back as `ErrorResponse`.
- Same async idiom: a quack table function that yields `BLOCKED` + an ASYNC-pool task, which is exactly what the read path will want once its FETCH goes async. The goal is one "async quack source" pattern reused in both directions.

## Open question: `QuackDataStream`

`QuackDataStream` is a small bounded queue with backpressure that bridges the request-handler threads (producers) to the table-function source (consumer). I expected the table-function async API to cover this, but it doesn't seem to: a table function gets no `InterruptState`, so it can't be woken from another thread the way `BufferedData`/`BatchMemoryManager` are, and `ManagedAsyncTaskQueue` is a task runner rather than a pull-based buffer. So I hand-rolled it. Is that the right call, or should something like this exist for table-function sources?

cc: @lnkuiper @pdet


src/quack_data_stream.cpp
```diff
@@ -0,0 +1,112 @@
...
+		std::lock_guard<std::mutex> guard(lock);
+		finished = true;
+	}
+	cv_nonempty.notify_all();
+	cv_nonfull.notify_all();
+}
+
+void QuackDataStream::SetError(ErrorData error_p) {
+	{
+		std::lock_guard<std::mutex> guard(lock);
```
lnkuiper: We should have `lock_guard` and `mutex` in the DuckDB namespace, and you could even use `annotated_lock_guard` and `annotated_mutex` with clang thread annotations (e.g., `DUCKDB_REQUIRES`) so the compiler can help finding locking issues

src/quack_server.cpp
```diff
@@ -398,43 +475,84 @@ unique_ptr<QuackMessage> QuackServer::HandleMessageInternal(DatabaseInstance &db
...
-		auto table_info = context.TableInfo(Identifier(append_request_message.SchemaName()),
-		                                    Identifier(append_request_message.TableName()));
-		if (!table_info) {
-			return make_uniq<ErrorResponse>("Table %s.%s does not exist",
-			                                SQLIdentifier(append_request_message.SchemaName()),
-			                                SQLIdentifier(append_request_message.TableName()));
+		// Lazily start the per-statement stream + its background INSERT on the first chunk. The INSERT
+		// runs `... SELECT * FROM scan_data_from_quack_client('<id>')`, engaging PhysicalBatchInsert.
+		// The create-once / lookup is done under the short-held lifecycle lock so concurrent producers
+		// (a future async client) can't double-create the stream or double-assign the thread.
```
lnkuiper: Maybe the comments could be a bit less verbose overall, lots of 4-line explanations

src/quack_server.cpp
```diff
@@ -398,43 +475,84 @@ unique_ptr<QuackMessage> QuackServer::HandleMessageInternal(DatabaseInstance &db
...
+				}
+				stream = QuackStreamRegistry::Get().Create(stream_id, send_data_message.AppendChunk().GetTypes());
+				connection.insert_stream = stream;
+				connection.insert_stream_id = stream_id;
+				connection.insert_thread = std::thread(RunInsertStatement, std::ref(connection), stream, stream_id,
+				                                       send_data_message.SchemaName(), send_data_message.TableName());
+			}
+		}
+		// Finalize any displaced stale stream without holding the lifecycle lock.
+		FinalizeDetached(displaced);
```
lnkuiper: This returns an `ErrorData` but seems to be ignored

lnkuiper (commented):
Really cool that we're using DuckDB's execution engine for inserts now! I have left a few comments below and Codex thinks it found a hang:
```sql
# name: test/temp.test
# description: local source failure after streamed insert chunks must not leave server insert hanging
# group: [temp]

include extension/external/quack/test/template/quack_setup.test_template

statement ok
attach {server} as rpc (token 'asdf')

statement ok
create table rpc.local_fail(i bigint);

statement error
insert into rpc.local_fail
select case when i >= 2048 then error('local source failed') else i end::bigint
from range(4096) t(i);
----
local source failed

statement ok
insert into rpc.local_fail values (-1);

query I
select count(*) from rpc.local_fail;
----
1

statement ok con2
detach rpc;

include extension/external/quack/test/template/quack_teardown.test_template

```

src/quack_server.cpp
guillesd: 100% agree

guillesd:
@lnkuiper addressed!

lnkuiper:
Looks great, thanks! One thing I'm curious about, this leaves `QUACK_VERSION` at `1`, but I don't think it's compatible anymore. Should we bump it to 2 and leave it there until the next quack release? Not urgent, you can do that in a follow-up PR.

PR #191, opened 2026-06-30
Async, order-preserving batch INSERTs

guillesd:
#189 moved the server-side append onto a real `INSERT INTO t SELECT * FROM scan_data_from_quack_client('<id>')` driven by `PhysicalBatchInsert`, but left the client deliberately dumb: it sent one chunk per `QUACK_SEND_DATA_REQUEST` and blocked the execution thread on each POST. That serializes the whole upload behind the network, and it throws away source order the moment more than one thread is producing.

This PR is the async/parallel client #189 said would come later, plus the server-side machinery needed to put the rows back in order. The upload now runs off the execution threads, multiple requests are in flight at once, and a remote `INSERT ... SELECT` lands rows in the same order it would locally. This supersedes https://github.com/duckdb/duckdb-quack/pull/179 and is also 2 to 2.5x faster in localhost.

This builds on #189 and keeps the read path untouched.

## Async upload

The sink no longer POSTs inline. A regular execution thread serializes its buffered chunks into a `QUACK_SEND_DATA_REQUEST` and registers the payload with a per-statement `ManagedAsyncTaskQueue`; an ASYNC-pool thread then does the blocking POST and checks the ack. So the producer threads keep scanning and serializing while the network work happens elsewhere, and several sends are outstanding at the same time.

To make that split clean I added `QuackClient::PostRaw`, which takes already-serialized bytes and returns the raw response body. The serialize step runs on the producer thread (where the `ClientContext` is live), and the POST runs on the async pool with `context=nullptr`, falling back to database-level params and logging. `FINALIZE` drains the queue before it returns, so an autocommit insert is still atomic on the server.

## Ordering modes

Order preservation is decided at plan time in `ConfigureOrdering`, mirroring core's `plan_insert.cpp`, and stored on the operator as an `AppendOrderMode`:

- `UNORDERED` (`preserve_insertion_order=false`): the fast path. No stamps, the server inserts chunks as they arrive.
- `PARALLEL_ORDERED`: parallel sources (table/parquet scans). Each thread stamps its chunks with `(batch_index, sequence_index, is_last_in_batch)` taken from the executor's batch index, and the server reassembles.
- `SERIAL_ORDERED`: single-threaded sources like `range()` that have no executor batch index. The lone producer mints a fresh batch per flush.

`RequiredPartitionInfo` only asks for executor batch indices in `PARALLEL_ORDERED`, so the executor's batch-index assertion never fires for sources that don't supply one. `ParallelSink` is true for everything except `SERIAL_ORDERED`.

## Wire protocol

`SEND_DATA_REQUEST` now carries a vector of chunks instead of a single one (one message is one flush), plus the ordering stamp and a `batch_watermark`. `FINALIZE` carries a `min_batch_watermark`.

The watermark is the piece that makes ordered delivery safe. It is the minimum batch index that will ever appear in the stream, piggybacked on every ordered message. Without it the server cannot tell whether a low batch is still coming or simply never existed, so it could deliver batch 5 before batch 4 ever arrives and trip `PhysicalBatchInsert`'s monotonicity check. With the watermark the server initialises its delivery cursor and only drains a batch once every batch below it is accounted for. `FINALIZE` repeats it as a safety net in case all messages arrived before any per-message watermark was valid.

## Server side: reassembly and parallel insert

`QuackDataStream` grew an ordered-assembly stage. `PushOrdered` buffers `(batch, seq)` fragments and, as soon as a batch's full sequence range has arrived, drains it as one unit into a `ready_batches_` queue in ascending batch order. Unordered streams skip all of that: `PushUnordered` drops one wire message's chunks straight onto the same queue.

On the scan side, `scan_data_from_quack_client` now raises `MaxThreads` to the live thread count. Each scan task claims one complete batch from `ready_batches_` via `TryPopBatch` and owns it exclusively until it is drained, so `PhysicalBatchInsert` sees a stable, unique batch index per task and can write row groups in parallel without interleaving rows across a batch boundary. When no batch is ready the task returns `BLOCKED` with an ASYNC-pool wait task, same as before.

Unordered streams take a different operator entirely: the bind signals `NO_ORDER`, which makes the planner pick `PhysicalInsert(parallel=true)` instead of `PhysicalBatchInsert`. That sink ignores batch indices, so concurrent tasks can append freely. Both modes share the one `ready_batches_` queue and the one scan loop; the only thing that differs is how the producer fills the queue and which insert operator drains it.

## Follow-ups

- `ready_batches_` is unbounded for now. Real flow control / backpressure management (the server handing the client an accept budget through the `SEND_DATA_RESPONSE` placeholder) is out of scope for this PR.
- Decoupling serialization from request (or a good abstraction that can be reused and does this) is a separate transport change that now is implemented partially for the purpose of async posting but could be reused in different parts (like the fetch path)
- Also memory in `QuackDataStream` is not really tracked by the buffer manager. Maybe we'd like this to be the case although I am not sure. Maybe just doing a good job on the backpressure side is enough. Or as Mark mentioned, make CollumnDataCollection serializable and use this in the transport (ColumnDataCollection is buffer managed)

### DIFF
--- a/src/include/quack_client.hpp
+++ b/src/include/quack_client.hpp
@@ -31,6 +31,15 @@ public:
 		return unique_ptr_cast<QuackMessage, TARGET>(std::move(response_message));
 	}
 
+	//! POST already-serialized request bytes and return the raw response body, throwing on transport failure.
+	//! Lets an async sender serialize on a producer thread and perform the blocking POST from the ASYNC pool;
+	//! pass context=nullptr when called off the execution thread (parameters fall back to the database).
+	virtual string PostRaw(optional_ptr<ClientContext> context, const_data_ptr_t data, idx_t size) = 0;
+
+	//! Emit a Quack request log entry. Safe to call from an async pool thread (uses db-level logger).
+	void LogRequest(MessageType request_type, const string &connection_id, optional_idx client_query_id,
+	                const string &query, int64_t duration_ms, MessageType response_type, const string &error);
+
 	static unique_ptr<QuackClient> GetClient(DatabaseInstance &db, const QuackUri &uri);
 	static unique_ptr<QuackClient> GetClient(ClientContext &context, const QuackUri &uri);
 
@@ -91,9 +100,13 @@ public:
 	HttpsQuackClient(DatabaseInstance &db, const QuackUri &uri_p);
 	~HttpsQuackClient() override;
 
+	string PostRaw(optional_ptr<ClientContext> context, const_data_ptr_t data, idx_t size) override;
+
 private:
 	unique_ptr<QuackMessage> RequestInternal(optional_ptr<ClientContext> context,
 	                                         unique_ptr<QuackMessage> request_message) override;
+	//! POST bytes assuming request_mutex is already held.
+	string PostRawLocked(const_data_ptr_t data, idx_t size);
 
 private:
 	unique_ptr<HTTPParams> http_params;
--- a/src/include/quack_data_stream.hpp
+++ b/src/include/quack_data_stream.hpp
@@ -9,73 +9,83 @@
 
 #include <condition_variable>
 #include <deque>
+#include <map>
 
 namespace duckdb {
 
-struct QuackStreamChunk {
-	unique_ptr<DataChunk> chunk;
-	//! Wire batch index (set by the future async/ordered client). Invalid for the current client,
-	//! in which case the scan assigns a constant batch index.
-	optional_idx batch_index;
-};
-
-//! Bounded, thread-safe queue feeding one server-side scan_data_from_quack_client INSERT. SEND_DATA
-//! handlers push chunks; the table function pops them; the bound backpressures the client.
+//! Thread-safe delivery queue for scan_data_from_quack_client. Producers push complete batches into
+//! ready_batches_; scan tasks claim one batch at a time via TryPopBatch for exclusive ownership.
 class QuackDataStream {
 public:
-	enum class PopStatus : uint8_t {
-		CHUNK,    //! a chunk was dequeued into `out`
-		EMPTY,    //! queue empty, more chunks may still arrive
-		FINISHED, //! queue empty and the stream is finished
-		ERRORED   //! the stream recorded a fatal error
+	enum class PopBatchStatus : uint8_t {
+		BATCH,
+		EMPTY,
+		FINISHED,
+		ERRORED,
 	};
 
-	explicit QuackDataStream(vector<LogicalType> types_p, idx_t capacity_p = 64)
-	    : types(std::move(types_p)), capacity(capacity_p) {
+	explicit QuackDataStream(vector<LogicalType> types_p, bool ordered_p = false)
+	    : types(std::move(types_p)), ordered(ordered_p) {
 	}
 
 	const vector<LogicalType> &Types() const {
 		return types;
 	}
 
-	//! Producer: enqueue a chunk, blocking while the queue is full (until the consumer drains or the
-	//! stream is finished/errored).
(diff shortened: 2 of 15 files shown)


lnkuiper:
Codex found this repro for when there are gaps in batch ids:
```sql
# name: test/temp.test
# description: ordered async quack insert must handle internal parquet batch-index gaps
# group: [temp]

require parquet

include extension/external/quack/test/template/quack_setup.test_template

statement ok
attach {server} as rpc (token 'asdf')

statement ok
SET threads=4;

statement ok
copy (select * from range(300000) t(i)) to '{TEST_DIR}/gap.parquet' (row_group_size 50000);

statement ok
create table rpc.gap(i bigint);

statement ok
insert into rpc.gap
select i from read_parquet('{TEST_DIR}/gap.parquet') where i < 4096 or (i >= 200000 and i < 204096);

query III
select count(*), min(i), max(i) from rpc.gap;
----
8192	0	204095

query I
select count(*) from (select i, row_number() over () - 1 as rn from rpc.gap) where (rn < 4096 and i != rn) or (rn >= 4096 and i != rn + 195904);
----
0

statement ok con2
detach rpc;

include extension/external/quack/test/template/quack_teardown.test_template
```

I also think `quack_message.json` is out of date with the source code? Seems like `serialize_quack_message.cpp` was edited manually.

guillesd:
@lnkuiper I fixed this via sending a dead markers range to the server, so we still keep the work on the server side without any client side buffering/dense remapping (which Mark didn't like as an idea since clients are usually smaller than servers)

lnkuiper:
There is an issue where if a batch with a higher batch index arrives earlier than a batch with a lower batch index, the lower batch index data is dropped:
```sql
# name: test/temp.test
# description: ordered async quack insert must not drop an earlier live batch when a later batch arrives first
# group: [temp]

require parquet

include extension/external/quack/test/template/quack_setup.test_template

statement ok
attach {server} as rpc (token 'asdf')

statement ok
SET threads=2;

statement ok
copy (
	select i, case when i < 50000 then md5(i::varchar) else 'x' end as s
	from range(300000) t(i)
) to '{TEST_DIR}/slow_leading.parquet' (row_group_size 50000);

statement ok
create table rpc.race(i bigint, l bigint);

statement ok
insert into rpc.race
select i, length(s)
from read_parquet('{TEST_DIR}/slow_leading.parquet')
where i < 4096 or (i >= 200000 and i < 204096);

query III
select count(*), min(i), max(i) from rpc.race;
----
8192	0	204095

query I
select count(*) from (select i, row_number() over () - 1 as rn from rpc.race)
where (rn < 4096 and i != rn) or (rn >= 4096 and i != rn + 195904);
----
0

statement ok con2
detach rpc;

include extension/external/quack/test/template/quack_teardown.test_template
```

There is also an issue with protocol versioning. Might not be related to the changes in this PR, and not necessary to fix in this PR, but I just wanted to add it because every time I ask Codex to find breakages it also brings this up:
> Medium: protocol negotiation still accepts incompatible clients. The server supports QUACK_VERSION = 2 (/private/tmp/duckdb-quack-pr189-duckdb/extension/external/quack/src/include/quack_server.hpp:96), but it only rejects clients whose minimum supported version is greater than 2 (/private/tmp/duckdb-quack-
pr189-duckdb/extension/external/quack/src/quack_server.cpp:393). A v1-only client with min=1,max=1 would still be accepted. The client also does not verify the response version before storing the connection id (/private/tmp/duckdb-quack-pr189-duckdb/extension/external/quack/src/quack_client.cpp:221). This
should check for range overlap on the server and validate the selected server version on the client.

lnkuiper:
Looks great! Ready to merge once CI passes

PR #194, opened 2026-07-02
Refactor/request internal primitives

guillesd:
Separate Encoding, Decoding and LogRequest from `RequestInternal` in `QuackClient`. We keep `RequestInternal` as our "sync" flow, but paths that need async machinery (like `SEND_DATA` and in a subsequent PR, `FETCH_DATA`), can decouple serializing from the actual post request. This enables a better decoupling of regular executor threads doing the CPU bound computation and async threads doing only IO.

Moving the logger also cleaned up a bunch of code.


lnkuiper:
Thanks!

PR #197, opened 2026-07-02
Fix internal error when projecting quack_query results that span multiple fetch batches

stephaniehingtgen:
**Overview**

Currently on main, queries against `quack_query(...)` throw an internal error when:
- the result is large enough to exceed one fetch batch (`quack_fetch_batch_chunks`, default 12)
AND
- DuckDB prunes the scan output to fewer columns than the inner query returns (e.g. an aggregate over one column of a two-column select)

You get an error like:
```
INTERNAL Error:
Attempted to access index 1 within vector of size 1
```

This PR fixes it by mirroring the `QuackScanInitGlobal` discriminator when tagging fetched chunks:
- `REQUIRES_PUSHDOWN` when `table_name` is empty (direct query)
- `PUSHDOWN_ALREADY_APPLIED` otherwise (catalog path)

**Root cause**

`QuackScan` tags each result chunk with whether projection still needs to be applied locally. For the initial batch, `QuackScanInitGlobal` does this by:
- on catalog/attach path (`table_name` set), rewriting the query with `BuildPushdownQuery`, so the server already projected (`PUSHDOWN_ALREADY_APPLIED`)
- on direct `quack_query('...')`, shipping the caller's SQL verbatim. the chunks come back full-width and need local projection (`REQUIRES_PUSHDOWN`).

The part I changed is the fetch path in `QuackScan`, where `PUSHDOWN_ALREADY_APPLIED` was hardcoded for every fetched chunk. For the direct-query path this made `DataChunk::Reference` wire a full-width chunk into the narrower pruned output, indexing past the end of the output's columns. That's why the first ~12 chunks scan fine and the error only fires once a chunk arrives via a fetch round-trip.

**Testing**
I added `test/sql/fetch_projection.test`, which creates 30000 rows with 2 columns. This forces a fetch round-trip past the 12-chunk batch limit, then aggregates over each column to verify both the width handling and that local projection picks the right column. It fails on main and passes with this change.

### DIFF
--- a/src/quack_scan.cpp
+++ b/src/quack_scan.cpp
@@ -332,9 +332,13 @@ static void QuackScan(ClientContext &context, TableFunctionInput &input, DataChu
 				bind_data.completed = true;
 				return;
 			}
-			// set up buffer for scan in next iteration
+			// tag fetched chunks like the initial batch (see QuackScanInitGlobal): direct queries
+			// return full-width chunks that still need projection, the catalog path already projected
+			auto fetched_pushdown_type = bind_data.table_name.empty()
+			                                 ? ChunkResultPushdownType::REQUIRES_PUSHDOWN
+			                                 : ChunkResultPushdownType::PUSHDOWN_ALREADY_APPLIED;
 			for (auto &chunk : fetch_response->MutableResults()) {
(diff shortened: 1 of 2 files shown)


lnkuiper:
Thanks!

Issue #114, opened 2026-05-12
INSERT RETURNING * returns phantom data


rustyconover:
### What happens?

`INSERT ... RETURNING *` against a table in an attached quack database returns phantom row data instead of the row that was actually inserted. The row is written to the remote server correctly — only the `RETURNING` output is wrong.

### To Reproduce

**Server** (DuckDB v1.5.2):

```sql
INSTALL quack FROM core_nightly;
LOAD quack;
CREATE TABLE inventory(id INTEGER PRIMARY KEY, name VARCHAR, qty INTEGER);
INSERT INTO inventory VALUES (1,'apple',10),(2,'banana',5),(3,'cherry',20);
CALL quack_serve('quack:localhost:9494', token='super_secret');
```

**Client** (separate DuckDB v1.5.2 instance):

```sql
INSTALL quack FROM core_nightly;
LOAD quack;
CREATE SECRET (TYPE quack, TOKEN 'super_secret');
ATTACH 'quack:localhost:9494' AS remote;

INSERT INTO remote.inventory VALUES (5, 'elderberry', 12) RETURNING *;
```

**Observed output:**

```
┌───────┬─────────┬───────┐
│  id   │  name   │  qty  │
│ int32 │ varchar │ int32 │
├───────┼─────────┼───────┤
│     1 │         │     0 │
└───────┴─────────┴───────┘
```

**Expected output:**

```
┌───────┬────────────┬───────┐
│  id   │    name    │  qty  │
│ int32 │  varchar   │ int32 │
├───────┼────────────┼───────┤
│     5 │ elderberry │    12 │
└───────┴────────────┴───────┘
```

A subsequent `SELECT * FROM remote.inventory` confirms the row `(5, 'elderberry', 12)` was inserted correctly on the server. Only the `RETURNING` projection is wrong — `id` defaults to `1`, `name` to `NULL`, `qty` to `0`, regardless of the values inserted.

### OS

macOS 14 (Darwin 24.6.0), arm64

### DuckDB Version

v1.5.2 (Variegata) 8a5851971f

### DuckDB Client

CLI

### Hardware

Apple Silicon

### Full Name

Rusty Conover

### Affiliation

Query.Farm

### What is the latest build you tested with? If possible, we recommend testing with the latest nightly build.

I have tested with a stable release

### Did you include all relevant data sets for reproducing the issue?

Yes

### Did you include all code required to reproduce the issue?

Yes, I have

[...]

PR #203, opened 2026-07-05
Fix INSERT ... RETURNING returning phantom data over quack

carlopi:
QuackInsert's source (GetDataInternal) always emitted a single BIGINT insert-count row, ignoring the RETURNING clause. For INSERT ... RETURNING, the operator's output types are the table's columns, so it wrote the count into column 0 and left the remaining columns uninitialized, producing phantom row data. The write path itself was correct.

Mirror upstream PhysicalInsert's return_chunk handling: capture the inserted chunks into a ColumnDataCollection when RETURNING is used and stream them back as the result; keep the count-only behavior otherwise.

Fixes upstream #114.


carlopi:
Replaced by https://github.com/duckdb/duckdb-quack/pull/219, needs proper rebase and all, for now closing.

PR #204, opened 2026-07-05
Add optional EXTRA_HTTP_HEADERS to the quack secret

carlopi:
The `quack` secret (TYPE quack) now accepts an optional EXTRA_HTTP_HEADERS map of VARCHAR->VARCHAR. Any headers provided are injected into every HTTP request the quack client sends to the server (connect/prepare/fetch/append/ disconnect all funnel through HttpsQuackClient::RequestInternal), which is useful for authenticating through a proxy or load balancer in front of the Quack server.

The headers are resolved once per client from the URI-scoped secret and reused for all subsequent requests. Adds a SQL test and README docs.

Note: this works explicitly for connection TO a quack endpoint.
An alternative, if you want to cover all traffic, is for example an explicit `HTTP` secret, see for example https://duckdb.org/docs/lts/core_extensions/httpfs/https#authenticating

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,19 @@ FROM remote.hello;
 
 This should show the content of the remote table `hello` on the client side.
 
+The `quack` secret also accepts an optional `EXTRA_HTTP_HEADERS` map. Any headers
+provided there are attached to every HTTP request the client sends to the server,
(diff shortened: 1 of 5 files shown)


Mytherin:
Thanks!

PR #205, opened 2026-07-05
Source quack_serve token from a matching secret when none is supplied

carlopi:
Previously quack_serve generated a random token whenever the `token` parameter was omitted, ignoring any quack secret scoped to the endpoint. The client path (QuackClient::ConnectToServer) already sourced its token from a matching secret; this gives the server the symmetric behavior.

Token resolution for quack_serve is now, in priority order:
  1. an explicit `token` parameter,
  2. the token of a quack secret matching the listen endpoint,
  3. a freshly generated random token.

The secret-lookup logic is factored into QuackServer::TokenFromSecret and reused by both the client and server paths. Adds test/sql/serve_token_from_secret.test.

This would allow for example to have a persistent secret on a given machine, and have all `quack_serve` created servers get it (with the relevant scope) from there, avoiding secrets.

### DIFF
--- a/src/quack_server.cpp
+++ b/src/quack_server.cpp
@@ -5,6 +5,7 @@
 #include "duckdb/main/client_context.hpp"
 #include "duckdb/main/connection.hpp"
 #include "duckdb/main/database.hpp"
+#include "duckdb/main/secret/secret_manager.hpp"
 #include "duckdb/parser/parsed_data/create_table_info.hpp"
 #include "duckdb/storage/temporary_file_manager.hpp"
 #include "duckdb/common/serializer/binary_deserializer.hpp"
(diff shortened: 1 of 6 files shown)


src/quack_server.cpp
```diff
@@ -32,6 +33,17 @@ QuackServer::QuackServer(ClientContext &context_p, const QuackUri &uri_p, const
 	ValidateToken(token);
 }
 
+string QuackServer::TokenFromSecret(ClientContext &context, const QuackUri &uri) {
+	auto &secret_manager = SecretManager::Get(context);
+	auto transaction = CatalogTransaction::GetSystemCatalogTransaction(context);
+	auto match = secret_manager.LookupSecret(transaction, uri.Uri(), "quack");
```
lnkuiper: Should this lookup use the canonical URI?

lnkuiper (commented):
Looks good, just one question:

src/quack_server.cpp
carlopi: That's correct, thanks!

carlopi:
This is ready on my side.

PR #207, opened 2026-07-07
Refactor server to elastic pool of workers

guillesd:
This PR removes the default pool of 128 worker threads on the server side and replaces it with an elastic pool. Amongst other things, removing this limit allows us to remove the cap on cached connections and avoid deadlocks when the number of concurrent connections reaches 128. Reusing sockets does improve transactional workloads in localhost up to 1.2x so I'm guessing in WAN this will be considerable.

This is a simpler alternative/straight forward fix that avoids re-engineering the server into a fool event loop driven machine.

PS: Don't mind the name of the branch, I was experimenting with event loops and ergo it's called like that but there's no event loop!

### DIFF
--- a/src/quack_client.cpp
+++ b/src/quack_client.cpp
@@ -1,7 +1,6 @@
 #include "duckdb/main/database.hpp"
 #include "duckdb/main/extension_helper.hpp"
 #include "duckdb/main/secret/secret_manager.hpp"
-#include "duckdb/parallel/task_scheduler.hpp"
 
 #include "quack_client.hpp"
 #include "quack_uri.hpp"
@@ -90,7 +89,7 @@ string HttpsQuackClient::PostRawLocked(const_data_ptr_t data, idx_t size) {
 	PostRequestInfo post_request(request_url, headers, *http_params, data, size);
 	unique_ptr<HTTPResponse> response;
 	try {
-		response = http_util.Request(post_request);
+		response = http_util.Request(post_request, http_client);
 	} catch (std::exception &ex) {
 		ErrorData error(ex);
 		throw IOException("Failed to send message: %s", error.Message());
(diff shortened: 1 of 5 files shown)


src/quack_client.cpp
```diff
@@ -235,12 +228,6 @@ unique_ptr<QuackClientWrapper> QuackClientConnection::GetClient(ClientContext &c
 
 void QuackClientConnection::StoreClient(unique_ptr<QuackClient> client_p) const {
 	lock_guard<mutex> guard(lock);
-	if (cached_clients.size() >= max_connections_cached) {
-		// already exceeded max cache size
-		return;
-	}
-	// A pooled client has no owning query; drop the stamp so later teardown doesn't log under it.
-	client_p->SetRequestLogger(nullptr);
 	cached_clients.push_back(std::move(client_p));
```
lnkuiper, line 248: This makes the client cache unbounded while each cached `HttpsQuackClient` now keeps a persistent HTTP socket. The insert path checks out a client before `ManagedAsyncTaskQueue::Register`/backpressure, so pending `SEND_DATA` tasks can own many more clients than the active async-thread count. Once those tasks complete, this line retains all of them as idle keep-alive sockets, letting one logical Quack connection consume the server’s `quack_server_max_connections` budget and causing later sends/finalize/attaches to be shed. Can we keep a bounded cache, or move client checkout into the async task so retained sockets are bounded by actual active send concurrency?

lnkuiper (commented):
Great idea! Almost ready to go, just one concern from Codex (also, there's a merge conflict now):

src/quack_client.cpp
guillesd: I re-added the bounded cache!

guillesd:
Should be good now @lnkuiper

PR #208, opened 2026-07-08
Async fetch read ahead

guillesd:
## Main intent

This PR implements the async fetch read ahead for the read path in Quack. It reuses most components from the async write path: most notably, an `QuackFetchAhead` wrapper on top of `ManagedAsyncQueue` that tops up `QuackFetchDataTask` up to a certain depth (which defaults to `async_threads` but can be altered by `quack_fetch_read_ahead`).

Once fetch responses come back, they are put into a `QuackDataStream`, which is the same component used in the server side scan (`quack_scan_from_cient.cpp`). This is just a queue of batches that pipeline executors can consume, acting as the interface between the async workers and the pipeline executors.

> I am aware that some of this components may benefit from re-using some things from core, particularly this open PR by Pedro https://github.com/duckdb/duckdb/pull/23662

## Other things
- Some cleanup of logger functions
- Defaulting to rows as a cap for the transport. This is the same that happens in send data and prevents filters from sending completely uneven batches and ending up with 1kb requests.

> We are not implementing a better cost model yet, which could be introduced when potentially implementing budgeting between server and client, and be able to apply backpressure in quack streams. This is particularly important for the server side insert path!

## Performance

We all want to see the numbers...

### Base performance
Fetch was already heavily parallelized, so there is no improvement (nor regression) on simple `SELECT * FROM lineitem` kind of queries. For sf=5, we have:

| topology | main | branch (read-ahead ON) | Δ |
|---|---:|---:|---:|
| local loopback | 2.42 s | 2.01 s | branch −17% |
| EC2↔EC2 (0.1 ms RTT) | 2.79 s | 2.73 s | branch −2% |
| mac→EC2 (20 ms RTT, home link) | 45.29 s | 44.95 s | branch −0.7% |

For EC2 <> EC2, which is the one that we probably care about the most, we are already saturating the network in both cases. There is a risky option to increase performance, increasing the `streaming_buffer_size` to make read ahead more effective. However, Claude tells me this is not buffer managed so in concurrent query scenarios you may have a very big buffer of non-spillable data.

### One example of a win that can be very nice multi-worker systems

If the client wants to do further execution locally, there is a very big benefit of decoupling IO workers from pipeline executors. If you limit the amount of executor threads, I've seen up to 10% improvement. But this is very query specific, so if I choose the right one I'm pretty sure I can see 2x. In fact I just did, running a `JOIN` with a client side table.

## Is there more?

I think besides the improvements in memory budgets, and buffer manage quack streams, there is the question of whether we want to improve performance over WAN. The only thing that would make a serious different here is probably compressing the payloads. This is probably pointless in big bandwidth cases (and maybe even detrimental), but could be optional for WAN and could make 4x type of improvements, since we are already saturating the network pretty well.

Also, after some profiling, I've found a considerable inefficency with the single cursor StreamingQueryResult to SimpleBufferData. The buffer can only be emptied once it reaches a threshold, and threads are constantly waking up to poll it. In core we could make SimpleBufferData return `CHUNK_READY` or similar when data arrives, then we could pool chunks immediately and the fetch ahead queue would be much more effective.


src/quack_scan.cpp
```diff
@@ -319,29 +326,52 @@ static void QuackScan(ClientContext &context, TableFunctionInput &input, DataChu
...
-		// if that did not work, we request more results
-		if (local_state.results.empty() && global_state.needs_more_fetch) {
-			auto &client = local_state.client_wrapper->GetClient();
-			auto fetch_response = client.Request<FetchResponseMessage>(
-			    context,
-			    make_uniq<FetchRequestMessage>(bind_data.client_connection->ConnectionId(), global_state.query_uuid));
-
-			if (fetch_response->MutableResults().empty()) {
+		// if that did not work, we consume the next prefetched batch from the read-ahead buffer
+		if (local_state.results.empty() && global_state.needs_more_fetch && global_state.fetch_ahead) {
```
Mytherin: The logic seems to be split between this file (`quack_scan.cpp`) and the `quack_fetch_ahead` class - maybe we can move this all there and call it `QuackFetcher` or so?

src/quack_scan.cpp
```diff
@@ -155,6 +156,8 @@ struct QuackScanGlobalState : GlobalTableFunctionState {
 	vector<idx_t> projection_ids;
 	atomic<bool> needs_more_fetch;
 	hugeint_t query_uuid;
+	//! FETCH read-ahead pipeline; set when needs_more_fetch (see QuackScanInitGlobal).
+	shared_ptr<QuackFetchAhead> fetch_ahead;
```
Mytherin: Maybe the `needs_more_fetch` can also be integrated into the fetch object?

I would just prefer to have the fetch logic live as much as possible in one place so it is easier to reason about.

src/quack_scan.cpp
```diff
@@ -319,29 +326,52 @@ static void QuackScan(ClientContext &context, TableFunctionInput &input, DataChu
...
-			    context,
-			    make_uniq<FetchRequestMessage>(bind_data.client_connection->ConnectionId(), global_state.query_uuid));
-
-			if (fetch_response->MutableResults().empty()) {
+		// if that did not work, we consume the next prefetched batch from the read-ahead buffer
+		if (local_state.results.empty() && global_state.needs_more_fetch && global_state.fetch_ahead) {
+			auto &fetch_ahead = *global_state.fetch_ahead;
+			idx_t batch_index;
+			vector<unique_ptr<DataChunk>> chunks;
+			switch (fetch_ahead.Buffer().TryPopBatch(batch_index, chunks)) {
```
Mytherin: Could you expand a bit on how we are guaranteeing these batches are coming out in-order here? It's not entirely clear to me why the below situation cannot occur:

```
REQUEST #1
REQUEST #2
REQUEST #3

RECEIVE #3
RETURN #3

RECEIVE #2
RETURN #2
```

Mytherin: Note that returning batches in anything but incrementing order is incorrect and can cause downstream issues / incorrect results / internal exceptions

Mytherin (commented):
Thanks for the PR! Looks good - can we also include a benchmark (see ducklake benchmark directory or duckdb benchmark directory for how this can be setup). Some comments below. Maybe @pdet can also have a look

src/quack_scan.cpp
Mytherin: Maybe we can add a debug setting that can help verify this behavior, that e.g. adds random sleeps to batch returns which will cause this to be triggered more frequently.

guillesd: 1. The `batch_index` is assigned server side and the cursor emits serial batches (so far so good)
2. The `batch_index` is sent over the wire.
3. The fetch task pushes the batch to an order stream `buffer->PushOrdered(chunks, batch_index, ...)`. This happens in `quack_fetch_ahead.cpp` under `ExecuteInternal`. 
4. Then in `quack_scan.cpp` we pop them from the ordered stream via `TryPopBatch`.

Because the `batch_index` is dense, it makes the watermark logic very simple on the fetch side. On the send side it was considerably harder.

There are a bunch of tests specifically for this. I can add the debug setting though, seems like a good idea to make sure the implementation is tight!

guillesd: The logic, that is mostly needed for the send path, lives in `QuackDataStream::DrainOrderedBatches()`

guillesd:
@Mytherin I refactored the fetching logic to be centralized in QuackFetcher and moved stuff out of the scan. I also added the debug setting!

Regarding the benchmarking infra, I think that could be a separate PR. I think it requires some thought, since running Quack benchmarks in localhost loopback doesn't really tell us much! For CI checks we could maybe think of mocking a server with delays in requests and a fake limited bandwidth

src/quack_scan.cpp
Mytherin: Could you maybe draw a diagram of what this looks like? Is the server waiting to send batches to ensure they arrive in serial order (would that then not make this much slower than necessary potentially?).

My correctness doubt is roughly this scenario:

```
SERVER
T1 - Send batch #1 
T2 - Send batch #10
T3 - Send batch #100

CLIENT
T1 - Receive batch #100
T1 - Emit batch #100
```

Does the server just not *start* sending batch 100 until batch 1/10 have been received? Or do we ship already and then separately ship watermarks?

guillesd: So first clarification: the batch_index assignation is dense on the server side. It was before this PR too. See https://github.com/guillesd/duckdb-quack/blob/feat/async-fetch-read-ahead/src/quack_server.cpp#L460 and https://github.com/guillesd/duckdb-quack/blob/feat/async-fetch-read-ahead/src/quack_server.cpp#L516. So on the fetch side there is no watermark needed. This is not the case on the send side (but we can discuss this in person if you want). This means that there's no wait needed on the server side, we grab one batch, stamp the index and send it over the wire, so they will be sent in order.

Then on the client side, your concern

```
SERVER
T1 - Send batch #1 
T2 - Send batch #2
T3 - Send batch #3

CLIENT
T1 - Receive batch #2 
T1 - PushOrdered(#2) → parked, want #1
T2 - Receive batch #3 
T2 - PushOrdered(#3) → parked, want #1 
T3 - Receive batch #1 
T3 - PushOrdered(#1) → drain releases #1, #2, #3
T3 - Scan pops #1, #2, #3 from the stream
```

So there's some buffering on the client side, but should be very minimal since the server side does send in order. I'm planning to implement backpressure in `QuackDataStream` via a budget communication between client and server for edge cases where results are too big (or we are very large data to the server).

guillesd:
Okey I'm bumping the submodule to get some stuff from Pedro's read ahead PR (mainly `ReadAheadJobCompletion`)

Mytherin:
WIP bump here - https://github.com/duckdb/duckdb-quack/pull/221

Mytherin:
Thanks!

PR #209, opened 2026-07-09
Fix pushdown type of fetched continuation batches for bind-time results (attached views, quack_query)

asubbarao:
## Problem

Scanning an ATTACH'd remote **view** (or any projected `quack_query` / `quack_query_by_name` result) fails with an internal error when the result spans more than one fetch batch and the local plan projects a subset of columns:

```
INTERNAL Error: Attempted to access index 1 within vector of size 1
```
(stack: `QuackScan` → `DataChunk::Reference`)

Minimal repro (also the new test): a 3-column view over `range(10_000_000)`, attached, then `SELECT count(*), sum(i) FROM rpc.big_view`. Small views work — the failure only appears once the result no longer fits the first prepare batch, so in practice it looks data-dependent (a view that worked yesterday crashes today after its underlying tables grow).

## Cause

`QuackScan` has two chunk modes (`ChunkResultPushdownType`):

- **Catalog table scans** build a server-side pushdown query in `QuackScanInitGlobal` (`bind_data.table_name` non-empty) → response chunks are already projection-width → `PUSHDOWN_ALREADY_APPLIED`, consumed via `output.Reference(response_chunk)`.
- **Bind-time results** (`quack_query`, `quack_query_by_name` — which is also how attached views are scanned, per `QuackViewCatalogEntry::CreateViewSQL`) are full-width → `REQUIRES_PUSHDOWN`, consumed by mapping `column_ids` primary indexes into the response chunk.

The fetch loop in `QuackScan`, however, tags **every continuation batch** as `PUSHDOWN_ALREADY_APPLIED` unconditionally. In the bind-time-result mode, batch 2+ is a full-width continuation of the bind-time result, so `output.Reference(full_width_chunk)` runs against a projection-narrow output chunk and throws — the reported index always equals the projected width. (When the projected width happens to equal the source width but the column order is permuted, the `Reference` would instead succeed silently with mis-ordered columns.)

## Fix

`QuackScanInitGlobal` records which mode the scan is in (`fetch_pushdown_type`: `PUSHDOWN_ALREADY_APPLIED` for the server-side pushdown path, `REQUIRES_PUSHDOWN` for bind-time results), and the fetch loop tags continuation chunks with that instead of the hardcoded value.

## Test

`test/sql/attach_view_fetch_projection.test` — projected aggregates over a 3-column, 100k-row attached view, plus the same through `quack_query`. It sets `quack_fetch_batch_chunks=1` so even a small view spans many FETCH batches, which keeps it fast enough for the default CI test run. Fails with the internal error when built from the pre-fix source (`INTERNAL Error: Attempted to access index 1 within vector of size 1`); passes with this change. `attach_views.test`, `attach.test`, and `attach_fetch_large.test_slow` still pass.

Fixes #199 — same root cause, different flavor: there the mis-tagged full-width continuation chunk dies at column 0 on the type check (`Vector::Reference used on vector of different type (source VARCHAR referenced BIGINT)`) before reaching the bounds error; the reported "works with LIMIT 1" matches (the result fits the first prepare batch, so no continuation fetch happens).

### DIFF
--- a/src/quack_scan.cpp
+++ b/src/quack_scan.cpp
@@ -144,10 +144,11 @@ struct QuackScanLocalState : public LocalTableFunctionState {
 
 struct QuackScanGlobalState : GlobalTableFunctionState {
 	explicit QuackScanGlobalState(vector<ColumnIndex> column_ids_p, vector<idx_t> projection_id_p,
-	                              vector<ChunkResult> results_p, bool needs_more_fetch_p, hugeint_t result_uuid_p)
+	                              vector<ChunkResult> results_p, bool needs_more_fetch_p, hugeint_t result_uuid_p,
+	                              ChunkResultPushdownType fetch_pushdown_type_p)
 	    : max_threads(needs_more_fetch_p ? MAX_THREADS : 1), column_ids(std::move(column_ids_p)),
 	      projection_ids(std::move(projection_id_p)), needs_more_fetch(needs_more_fetch_p), result_uuid(result_uuid_p),
-	      results(std::move(results_p)) {
+	      fetch_pushdown_type(fetch_pushdown_type_p), results(std::move(results_p)) {
 	}
 	idx_t MaxThreads() const override {
 		return max_threads;
@@ -157,6 +158,12 @@ struct QuackScanGlobalState : GlobalTableFunctionState {
 	vector<idx_t> projection_ids;
 	atomic<bool> needs_more_fetch;
 	hugeint_t result_uuid;
+	//! How chunks fetched AFTER the initial batch must be treated. When the scan was
+	//! executed via a server-side pushdown query (catalog table scan), fetched chunks
+	//! already have the projection applied. When the scan consumes a bind-time result
+	//! (quack_query / quack_query_by_name, incl. ATTACH'd views), fetched chunks are
+	//! full-width continuations and still require client-side pushdown.
+	ChunkResultPushdownType fetch_pushdown_type;
 
 	vector<ChunkResult> TryGetResults() {
 		lock_guard<mutex> guard(lock);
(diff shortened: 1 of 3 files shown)


asubbarao:
### Reviewer-facing proof: graded isolation, pre-fix vs patched build

To make the fail→pass concrete, here's the same query run against **two real client builds** on one live server — `buggy` = stock pre-patch extension, `fixed` = this branch. Each case removes exactly one of the three conditions the bug needs (bind-time result **+** column projection **+** a continuation fetch batch); only the case where all three hold crashes, and only on the pre-fix build.

Server-side objects: a wide **view** and wide **table** (100k rows each → spans >1 fetch batch) and a small **view** (100 rows → single batch), all `SELECT i AS a, i AS b`.

```
ATTACH 'quack://…' AS <db> (TYPE quack, token '…');

(a) multibatch VIEW, NO projection      SELECT sum(a)+sum(b) FROM <db>._bug_demo_wide;   -- reads both cols, no narrowing
      fixed:  ✓ 9999900000     buggy:  ✓ 9999900000

(b) projected VIEW, SINGLE batch        SELECT sum(a) FROM <db>._bug_demo_small;         -- narrows to {a}, 100 rows = 1 batch
      fixed:  ✓ 4950           buggy:  ✓ 4950

(c) projected VIEW, MULTI-batch         SELECT sum(a) FROM <db>._bug_demo_wide;          -- narrows to {a}, 100k rows, VIEW = bind-time
      fixed:  ✓ 4999950000     buggy:  ✗ INTERNAL Error: Attempted to access index 1 within vector of size 1

(d) projected TABLE, MULTI-batch        SELECT sum(a) FROM <db>._bug_demo_tbl;           -- identical to (c) but TABLE = server pushdown
      fixed:  ✓ 4999950000     buggy:  ✓ 4999950000
```

`(c)` vs `(d)` is the crux and maps directly to the `bind_data.table_name.empty()` branch this PR keys on: identical `SELECT sum(a)`, identical 100k rows — only view→table differs. The **table** scan sets `table_name`, so `BuildPushdownQuery` sends `SELECT a FROM …` and the server returns already-narrow chunks (`PUSHDOWN_ALREADY_APPLIED`); the **view** is a bind-time result whose continuation batches are full-width and were mis-tagged. `(a)`/`(b)` confirm the projection and the second fetch are each independently necessary.

carlopi:
Thanks a ton for the contriubution, I though of improving it as https://github.com/duckdb/duckdb-quack/pull/216 (where adding the same scheme also for not yet handled usages of `ChunkResultPushdownType::PUSHDOWN_ALREADY_APPLIED`, see for example:
```
		for (auto &chunk_ref : response_message->MutableResults()) {
			auto &chunk = chunk_ref->Chunk();
			results.emplace_back(chunk, ChunkResultPushdownType::PUSHDOWN_ALREADY_APPLIED);
		}
```
That currently is unreachable, but I could see future code changes bringing the bug back.

Adapted and expanded the tests.

PR #210, opened 2026-07-10
Implement catalog enumeration for attached databases (duckdb_tables, SHOW TABLES, IDE introspection)

asubbarao:
## Problem

Attached quack databases are invisible to catalog enumeration. With `ATTACH 'quack:...' AS rpc`:

- `duckdb_tables()` / `duckdb_views()` / `duckdb_columns()` return no rows for `rpc`
- `SHOW TABLES` (with `rpc` as the default database) shows nothing
- JDBC `DatabaseMetaData.getTables()` returns nothing, so IDEs and SQL tools (DataGrip, DBeaver, Harlequin, ...) that introspect via the standard catalog show an empty database — no table listing, no autocomplete — even though direct queries against `rpc.<table>` work fine.

## Cause

`QuackSchemaCatalogEntry::Scan` (both overloads) is a `// TODO` stub. The information is already there: at ATTACH time, `QuackTableSet` eagerly constructs bound catalog entries for every remote table and view (from the `QuackLoadCatalogData` bulk load), and `QuackCatalogSet::GetAllCatalogEntries()` exposes them. Name lookups (`LookupEntry`) work; only enumeration was left unimplemented.

## Fix

Implement `Scan` as an iterate-and-filter over the already-loaded table set: emit entries matching the requested `CatalogType` (`TABLE_ENTRY` / `VIEW_ENTRY`), nothing for other types. No new RPCs — the entries were loaded at attach. The context-taking overload delegates to the context-free one.

## Test

`test/sql/attach_catalog_scan.test` — attaches a served database with two tables and a view, then asserts `duckdb_tables()`, `duckdb_views()`, `duckdb_columns()` and `SHOW TABLES` all enumerate them. Fails on the current branch tip (all enumerations empty); passes with this change.

### DIFF
--- a/src/storage/quack_schema.cpp
+++ b/src/storage/quack_schema.cpp
@@ -68,10 +68,17 @@ optional_ptr<CatalogEntry> QuackSchemaCatalogEntry::LookupEntry(CatalogTransacti
 
 void QuackSchemaCatalogEntry::Scan(ClientContext &context, CatalogType type,
                                    const std::function<void(CatalogEntry &)> &callback) {
-	// TODO
+	Scan(type, callback);
 }
 void QuackSchemaCatalogEntry::Scan(CatalogType type, const std::function<void(CatalogEntry &)> &callback) {
-	// TODO
+	if (type != CatalogType::TABLE_ENTRY && type != CatalogType::VIEW_ENTRY) {
+		return;
+	}
+	for (auto &entry : tables->GetAllCatalogEntries()) {
+		if (entry.get().type == type) {
+			callback(entry.get());
+		}
+	}
 }
 
 optional_ptr<CatalogEntry> QuackSchemaCatalogEntry::CreateIndex(CatalogTransaction transaction, CreateIndexInfo &info,
--- a/src/storage/quack_table.cpp
+++ b/src/storage/quack_table.cpp
@@ -69,9 +69,11 @@ string QuackTableSet::GetLoadQuery() {
 	return R"(
 SELECT schema_name, sql, 'table'
 FROM duckdb_tables()
+WHERE NOT internal
 UNION ALL
 SELECT schema_name, view_name, 'view'
 FROM duckdb_views()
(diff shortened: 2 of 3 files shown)


asubbarao:
Closing — `QuackSchemaCatalogEntry::Scan` is implemented on main now (with the `SCHEMA_ENTRY` case this PR didn't cover), and `GetStorageInfo` returns an empty `TableStorageInfo`, so the enumeration half of this is done. The branch is also stale against the `schema_oid` / nested-schema rewrite of `GetLoadQuery`.

One piece here didn't land, and implementing `Scan` is what makes it visible: `QuackTableSet::GetLoadQuery()` still selects from `duckdb_tables()`/`duckdb_views()` without `WHERE NOT internal`. A bare served database reports 47 internal views (22 `pg_catalog`, 11 `information_schema`, 14 in `main` — including `duckdb_tables`, `duckdb_views`, `duckdb_columns`, `sqlite_master`), and they really are loaded into the client catalog today: `FROM r.sqlite_master` resolves over an ATTACH. They were harmless while `Scan` was a stub; now they'll enumerate through `SHOW TABLES` and IDE introspection. Happy to send that one-line filter plus a test as a focused PR.

asubbarao:
Correction to my closing comment: the `WHERE NOT internal` part is **not** needed on main, and I was wrong to flag it.

I built main (`9823b72`) and checked. The internal views live in the `system` database, not in the served user database — I had grouped only by `schema_name` and misread `system.main` as the user's `main`. They leaked into the attached catalog on older builds only because `GetLoadQuery` matched schemas by *name*, so `system.main` collided with the client's `main`. The `schema_oid` rewrite in 15ab79bf removed that as a side effect.

Verified on a main build: attaching a served database with one table and one view enumerates exactly `[t]` and `[v]`, and `r.sqlite_master` no longer resolves. Nothing left over from this PR — it's fully superseded. Sorry for the noise.

PR #211, opened 2026-07-11
Deduplicate quack scan output column names

carlopi:
A remote query is free to produce duplicate column names (SELECT 1 AS x, 2 AS x), but a table function binding requires them to be unique: duckdb's Binding::Initialize throws 'table "quack_query" has duplicate column name'. Both quack scan binds passed the server's names through verbatim, so any such query failed to bind.

Rename repeats in the bind (x, x_1, ...) via QueryResult::DeduplicateColumns, as the arrow scan does. Nothing maps back: quack never sends column names to the server - projection pushdown emits positional #N refs and the scan references result columns by index - so this only affects the names exposed to the binder.

Currently:
```sql
CALL quack_serve('quack:localhost', token:='1234');
SELECT * FROM quack_query('quack:localhost', 'SELECT 1 AS x, 2 AS x');
```
fails like:
```
Binder Error:
table "quack_query" has duplicate column name "x"
```
Failure happens client side, basically this goes local duckdb -> quack protocol -> computed server side -> quack protocol -> fails re-entering client side.

This is not great, since query has been executed (so possibly with side-effects) but then fails on re-entering client side.

After this PR:
```
┌───────┬───────┐
│   x   │  x_1  │
│ int32 │ int32 │
├───────┼───────┤
│     1 │     2 │
└───────┴───────┘
```
that behave according to the precedent of:
```sql
FROM (SELECT 1 AS x, 2 AS x);
```

Given this expand what's processed, and changes nothing for already de-duplicated queries, this is in my view an improvement.

Discovered while stress testing `CONNECT`, but found `USE quack` had the same limitation.

### DIFF
--- a/src/quack_scan.cpp
+++ b/src/quack_scan.cpp
@@ -1,5 +1,6 @@
 #include "duckdb/function/table_function.hpp"
 #include "duckdb/main/database.hpp"
+#include "duckdb/main/query_result.hpp"
 #include "duckdb/main/extension_helper.hpp"
 #include "duckdb/main/secret/secret.hpp"
 #include "duckdb/planner/filter/conjunction_filter.hpp"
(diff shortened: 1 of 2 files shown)


Mytherin:
Thanks!

PR #212, opened 2026-07-11
Quack: surface actual error types (instead of generic `Invalid`) AND allow invalidated servers to communicate this back to clients

carlopi:
This started looking to improve compatibility of `CONNECT` statement, and I bumped against the fact queries that would error server side (for N reasons, like `Catalog Errors`, constraints, Invalid Input, IOExceptions) would all surface as Invalid Input Exception, basically loosing information on the wire.

Idea here is keeping information, allowing more faith-full error messages to users.

Then I bumped against a corner case: `FatalException` would invalidate server AND client. That's not great, so for that and Internal error case I still opted for generic like message (but carrying around metadata).
Next: I realized that `FatalException` or in any case server being invalidated must be communicated back to client directly, since otherwise any other reconnection attempt would fail even at executing auth / authz queries.

Now added a boolean on ErrorMessage, so client can be informed and act on server in invalid state (example, earlier failures).

### DIFF
--- a/src/include/quack_message.json
+++ b/src/include/quack_message.json
@@ -32,9 +32,27 @@
         "name": "message",
         "type": "string",
         "serialize_property": "error.RawMessage()"
+      },
+      {
+        "id": 2,
+        "name": "exception_type",
+        "type": "string",
+        "serialize_property": "ExceptionTypeName()"
+      },
+      {
+        "id": 3,
+        "name": "extra_info",
+        "type": "unordered_map<string, string>",
+        "serialize_property": "TransferableExtraInfo()"
+      },
+      {
+        "id": 4,
+        "name": "must_invalidate",
+        "type": "bool"
       }
     ],
-    "constructor": ["message"]
+    "constructor": ["message", "exception_type", "extra_info", "must_invalidate"],
+    "constructor_method": "ErrorResponse::FromWire"
   },
   {
     "class": "PrepareRequestMessage",
--- a/src/include/quack_client.hpp
+++ b/src/include/quack_client.hpp
@@ -29,12 +29,22 @@ public:
 		if (response_message->Type() != TARGET::TYPE) {
 			if (response_message->Type() == MessageType::ERROR_RESPONSE) {
 				// if we get an error throw it immediately
+				NoteError(response_message->Cast<ErrorResponse>());
 				response_message->Cast<ErrorResponse>().Error().Throw();
 			}
 			throw IOException("Expected %s message, got %s instead", MessageTypeToString(TARGET::TYPE),
 			                  MessageTypeToString(response_message->Type()));
 		}
 		return unique_ptr_cast<QuackMessage, TARGET>(std::move(response_message));
+	}
+
+	//! Invalidate the owning connection if the error says the server's database died. Call before rethrowing.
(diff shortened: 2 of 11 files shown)


src/include/quack_message.json
```diff
@@ -32,9 +32,27 @@
         "name": "message",
         "type": "string",
         "serialize_property": "error.RawMessage()"
+      },
+      {
+        "id": 2,
+        "name": "exception_type",
+        "type": "string",
+        "serialize_property": "ExceptionTypeName()"
```
Mytherin, line 40: Maybe we can transfer the JSON representation instead? (`ConvertErrorToJSON`)

carlopi: This works for well formed errors, but it's not really closed in a bunch of weird cases, like, `raw_message` can start with `{` and not be JSON (for example, the weird looking `SELECT error('{aaa');`).

Due to:
```c++
void ErrorData::ConvertErrorToJSON() {
        if (!raw_message.empty() && raw_message[0] == '{') {
                // empty or already JSON
                return;
        }
        raw_message = StringUtil::ExceptionToJSONMap(type, raw_message, extra_info);
        final_message = raw_message;
}
```

I am not sure if solving that would just be better then.

sounkou-bioinfo:
This PR addresses a concrete downstream need in [CanardAbsurd](https://github.com/RGenomicsETL/CanardAbsurd), a DuckDB workflow state machine with independent R workers. I found this existing PR while checking related reports, so I am adding the use case here rather than opening a duplicate issue.

Our claims, checkpoints, heartbeats, completions, and cancellation are single SQL statements contending on a task row. Handlers run outside transactions and can have external effects. We need to distinguish a confirmed aborted operation from an ambiguous transport failure, and let the caller decide whether to retry **that statement**, not rerun the handler. An I/O error or lost response must not be treated as proof that a write did not commit.

With the currently installed community Quack build (`1693647` on DuckDB 1.5.3), I reproduced a conflicting remote task update reaching R as `INVALID_INPUT`, with the conflict identifiable only in the displayed message. The current main-branch codec still serializes `error.RawMessage()` alone, and deserialization reconstructs an `INVALID_INPUT` error. Preserving the original type and transferable metadata is preferable to classifying that text.

Similar needs arise in ETL/ingestion clients, idempotent submission and deduplication, connection-pool recovery, cancellation handling, and error telemetry. A broad error type is not by itself a complete retry policy: distinguishing an expected duplicate key from an unrelated constraint violation may need more specific metadata, and an invalidated remote server must not invalidate the local client process indiscriminately.

There is also a separate R-side loss: the low-level driver constructs a `duckdb_error` with fields, but its caller-oriented DBI rethrow drops them and the parent condition. I opened [duckdb/duckdb-r#2711](https://github.com/duckdb/duckdb-r/issues/2711) with a reproduction on R package versions 1.5.3 and 1.5.5. Both boundaries need to preserve the information for R callers to use it.

Useful coverage for our case would be transaction conflicts and constraint errors through `quack_query()`, including prepare/execute/fetch paths, plus mixed-version peers where missing or unknown metadata remains explicitly unknown. Transport failures should remain distinct from server-reported query failures. I have reviewed this PR's description and diff, but have not tested its branch.

---
Prepared with GPT 6 Astra (xhigh thinking) in pi coding agent.

PR #215, opened 2026-07-12
Drain the catalog-load query across fetch batches

carlopi:
QuackCatalog::ExecuteCommandInternal appended only the chunks carried by the PREPARE response and never issued a FETCH, so anything past the first batch was dropped. Its own comment said so: "FIXME this will break with many results!". For these queries the dropped rows are not a truncated result set - they are catalog entries, so a schema, table or view silently ceases to exist, with no error at all.

Reachable today: the table-load query is duckdb_tables() UNION ALL duckdb_views(), whose two branches emit separate chunks, so a database holding both a table and a view already spans two chunks and loses the view at quack_fetch_batch_chunks=1. At the default batch size it takes a large catalog to reach, but lowering that setting - a documented knob - silently deletes part of the catalog.

Fetch until the server reports the result exhausted, the same way QuackScan does.

### DIFF
--- a/src/storage/quack_catalog.cpp
+++ b/src/storage/quack_catalog.cpp
@@ -66,7 +66,6 @@ const QuackUri &QuackCatalog::GetServerUri() {
 }
 
 unique_ptr<ColumnDataCollection> QuackCatalog::ExecuteCommandInternal(ClientContext &context, const string &query) {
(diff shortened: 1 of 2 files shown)


guillesd:
Yes, it was about time we fixed this!

Mytherin:
Thanks!

PR #216, opened 2026-07-13
Fix pushdown with pruned columns via tag predicate

### LINKED ISSUE #146: quack with zipfs extension invalidates client database on second run
```
ATTACH 'quack:localhost' AS remote ;

FROM remote.query("LOAD zipfs; FROM read_csv('zip://https://icem7.fr/data/donnees_communes2022.zip/donnees_communes2022.csv')");
```
-- this works fine

then  again 
`FROM remote.query("FROM read_csv('zip://https://icem7.fr/data/donnees_communes2022.zip/donnees_communes2022.csv')");`

=> 
```
Invalid Input Error: Attempting to execute an unsuccessful or closed pending query result
Error: FATAL Error: Failed: database has been invalidated because of a previous fatal error. The database must be restarted prior to being used again.
Original error: "Attempted to access index 1 within vector of size 1"
```

carlopi:
Solves a bug that was solved by three independent PRs:
* https://github.com/duckdb/duckdb-quack/pull/209
* https://github.com/duckdb/duckdb-quack/pull/199
* https://github.com/duckdb/duckdb-quack/pull/197 (vs `main`)
and fixes https://github.com/duckdb/duckdb-quack/issues/146

When multiple roundtrips are required for fetching a result AND some columns are pruned client side from the output, the bug would surface either as:
```
INTERNAL Error:
Attempted to access index 1 within vector of size 1
```
or
```
INTERNAL Error:
Vector::Reference used on vector of different type (source VARCHAR referenced BIGINT)
```

All three PRs already solved the problem in the strict sense, here though it generalize the solution and make the invariant clear also in code-paths there are at the moment unreachable, but could if code expands re-surface the same issue.

Infrastructure from https://github.com/duckdb/duckdb-quack/pull/209 (this PR is on top of tht), adding extended testing, and moving a pre-existing `.test_slow` to `.test` given it's regularly sub 1 second.

### DIFF
--- a/src/quack_scan.cpp
+++ b/src/quack_scan.cpp
@@ -158,11 +158,12 @@ struct QuackScanGlobalState : GlobalTableFunctionState {
 	vector<idx_t> projection_ids;
 	atomic<bool> needs_more_fetch;
 	hugeint_t result_uuid;
-	//! How chunks fetched AFTER the initial batch must be treated. When the scan was
-	//! executed via a server-side pushdown query (catalog table scan), fetched chunks
-	//! already have the projection applied. When the scan consumes a bind-time result
-	//! (quack_query / quack_query_by_name, incl. ATTACH'd views), fetched chunks are
-	//! full-width continuations and still require client-side pushdown.
+	//! How EVERY chunk of this scan must be treated - the initial batch and all fetched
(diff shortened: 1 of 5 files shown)


Mytherin:
Thanks!

PR #217, opened 2026-07-13
QuackGetVirtualColumns to return only COLUMN_IDENTIFIER_EMPTY and not having partial (and potentially buggy) support for rowid

carlopi:
This is 2 commits stacked on top of https://github.com/duckdb/duckdb-quack/pull/216 (to be reviewed first)

What this does is take extra comments in the PR comment of https://github.com/duckdb/duckdb-quack/pull/199, take inspiration from work from @akhenakh in PR 199, thanks a ton, and clean it up by using `continue` vs `return` (not really a triggerable bug, but ready to be unlocked by other changes) AND changes from declaring `ROW_ID` as supported but wrong, to use the proper `EMPTY` virtual column, and throwing on unsupported virtual columns, so that path for expansion becomes clear.

This has client only changes, and no serialization changes, so still good for the v1.5-variegata branch with QUACK_VERSION = 1.

### DIFF
--- a/src/quack_scan.cpp
+++ b/src/quack_scan.cpp
@@ -185,8 +185,12 @@ static string BuildPushdownQuery(const QuackScanBindData &bind_data, const Table
 			}
 			if (col_id.IsVirtualColumn()) {
 				auto virtual_column = col_id.GetPrimaryIndex();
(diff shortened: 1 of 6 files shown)


Mytherin:
Thanks!

PR #218, opened 2026-07-14
Quack compression or make quack faster over WAN!

guillesd:
Builds on top of https://github.com/duckdb/duckdb-quack/pull/208 (assuming that will get merged).

This PR adds optional zstd compression to the two bulk transfer paths: FETCH/PREPARE responses (read) and SEND_DATA requests (append). Compressed HTTP bodies get an 11-byte frame (magic + codec byte + uncompressed size) and receivers auto-detect it, so client and server settings are fully independent and mismatched configurations just work. Control messages are never compressed, so the handshake can't break. zstd comes from duckdb's vendored copy, no new dependencies.

Server response compression is connection scoped rather than global. Each FETCH/PREPARE response is compressed according to the connection's server-side session value, falling back to the server's `SET GLOBAL` as the fleet default. This lets one server serve a same-region client raw and a WAN client compressed at the same time.

## User API

```sql
-- attach: compresses this attach's uploads and its responses from the server
ATTACH 'quack:host:port' AS rpc (token '...', compression 'zstd-1');

-- quack_query: scoped to that call's private connection
SELECT * FROM quack_query('quack:host:port', 'from t', token='...', compression='zstd-1');

-- underlying settings (sender-side, on whichever node they are set)
SET quack_compression = 'none' | 'zstd' | 'zstd-1' .. 'zstd-22';  -- default zstd = level 3
SET quack_compression_min_size = 4096;                            -- below this, send raw
```

The default is `'none'`.

## Numbers

zstd-1 compresses lineitem ~3.7x on the wire. WAN (laptop wifi to EC2, SF5): **reads go from 42.6s to 14.5s (2.9x)**, appends from **23.2s to 6.7s (3.5x**, uplink fully saturated). EC2 to EC2 same region (c7g.2xlarge, 14.9 Gbps): appends are slightly faster with zstd-1 (0.96s to 0.90s), reads slightly slower (3.06s to 3.31s) but this is due to a bottleneck on the server side serial batch production (that may be addressed later on in DuckDB core).

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -5,6 +5,7 @@ add_library(
   quack_active_connections.cpp
   quack_clear_cache.cpp
   quack_client.cpp
+  quack_compression.cpp
   quack_data_stream.cpp
   quack_extension.cpp
   quack_fetch_ahead.cpp
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -7,6 +7,7 @@ add_library(
   quack_client.cpp
   quack_data_stream.cpp
   quack_extension.cpp
+  quack_fetch_ahead.cpp
   quack_http_server.cpp
   quack_log.cpp
   quack_message.cpp
--- a/src/include/quack_compression.hpp
+++ b/src/include/quack_compression.hpp
@@ -1,0 +1,39 @@
+#pragma once
+
+#include "duckdb/common/allocator.hpp"
+#include "duckdb/common/serializer/memory_stream.hpp"
+
+namespace duckdb {
+
+class ClientContext;
+class DatabaseInstance;
+
+enum class QuackCodec : uint8_t { NONE = 0, ZSTD = 2 };
+
+//! Sender-side settings; receivers auto-detect from the frame header.
+struct QuackCompressionConfig {
+	QuackCodec codec = QuackCodec::NONE;
+	int64_t zstd_level = 3;
+	idx_t min_size = 4096;
+
+	static QuackCompressionConfig FromContext(ClientContext &context);
+	//! Parse 'none', 'zstd' or 'zstd-<1..22>'; throws on anything else.
+	void ParseSpec(const string &spec);
+};
+
+//! Frame for compressed HTTP bodies: magic (2) + codec (1) + uncompressed size (8) + codec stream.
+//! Raw bodies always have 0x00 as their second byte (BinarySerializer field id), so no collision.
+struct QuackCompression {
+	static constexpr uint8_t MAGIC0 = 0xDB;
+	static constexpr uint8_t MAGIC1 = 0xC2;
+	static constexpr idx_t FRAME_HEADER_SIZE = 11;
+	static constexpr idx_t MAX_DECOMPRESSED_SIZE = idx_t(1) << 32;
+
+	//! Returns false (out untouched) when codec is NONE, size < min_size, or compression doesn't shrink.
+	static bool Compress(const QuackCompressionConfig &config, const_data_ptr_t data, idx_t size, MemoryStream &out);
+	static bool IsCompressedFrame(const_data_ptr_t data, idx_t size);
+	//! Throws IOException on unknown codec, corrupt frame or cap violation.
+	static AllocatedData Decompress(const_data_ptr_t data, idx_t size, idx_t &uncompressed_size);
+};
+
+} // namespace duckdb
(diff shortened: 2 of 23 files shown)


carlopi:
One question, should the compression be added like this or done via HTTP headers? Like, given the stack is fully HTTP, I would consider piggybacking on the accept compression (possibly even pluggable then, like `gzip` or `br`), instead of making this part of the protocol?

carlopi:
And I guess another question: could this be negotiated as part of the protocol / ATTACH handshake? Instead of requiring manual opt-in.

guillesd:
Hi @carlopi!

1) I did not include this (maybe I should have) but I included the default gzip compression done via http headers in a the benchmark I run and it was incredibly slow. 10 to 30x concretely.

2) I'm not sure what you mean. There is already an ATTACH option. The nice thing (in my opinion) is that you may have:
- WAN connection that `ATTACH 'quack:host:port' AS remote (token '...', compression 'zstd-1');` and benefits from compression in all queries run under this quack connection.
- Another concurrent connection over EC2 that decides to opt-out.

I personally think this is what's cute about the design, but maybe I'm missing something! Let me know.

carlopi:
I think indeed it would need to be implemented.

I was thinking as protocol negotiation the following, taking inspiration from HTTP compression:
in the `ConnectMessage` (or however is called the first message to create a `Connection`), have client send "I support this list of compression methods" + serializing a `vector<string>` (maybe ordered by preference), and have the server take that list, take his own list, and figure out which to choose, and send back "this connection should use compression method: gzip" or "this connection should use compression method: identity".

I think adding negotiation of params is also needed going into stability, to allow protocol to be stable (so v2.0 talks to anyone), but allow v2.4 server and v2.3 client (for example) to negotiate what's the actual settings to use for the length of the connection.

I think allowing user opt-in is also great to have, but I think the ideal path is for this to be by default auto-detected by the 2 duckdb, that have all relevant info (could also take into account ping time or whatever to decide)

guillesd:
okey I think we should: allow the user to decide, but Indeed have some sort of negotiation: where the server says I accept this compressions and the client agrees and viceversa.

Yes I am convinced, compression_codec should be a header actually, not inferred from the payload.

Mytherin:
This is a cool idea but having this as an attach parameter is imo not the right place for it. The best form for this I think is to do this dynamically as we stream large amounts of data, i.e. we could "gradually" try to move one way or the other based on how fast transfer is going and dynamically adjust (also within a single result set transfer). It's not that expensive to compress a single message with `ZSTD-1`. If that ends up faster, we can try `ZSTD-2`, if not we back off to uncompressed, etc.

We can keep the attach operator (or some setting) as an override for the dynamic behavior, but I think the dynamic behavior will be much nicer for users given that this is connection / latency / CPU dependent.

Mytherin:
Quack messages already have a `MessageHeader`. That header could hold how the rest of the payload is compressed (or not) on a per message basis.

Mytherin:
> I think adding negotiation of params is also needed going into stability, to allow protocol to be stable (so v2.0 talks to anyone), but allow v2.4 server and v2.3 client (for example) to negotiate what's the actual settings to use for the length of the connection.

Wouldn't that just be negotiating a version? i.e. v2.4 and v2.3 negotiate and use v2.3 - with all that entails (e.g. supporting compression XYZ, etc.)

guillesd:
I like the idea of the `MessageHeader` with the compression codec. The dynamic compression depending on link speed could be a follow up I think.

carlopi:
> Wouldn't that just be negotiating a version? i.e. v2.4 and v2.3 negotiate and use v2.3 - with all that entails (e.g. supporting compression XYZ, etc.)

I think advantage of scheme like:
-> client provide info / preferences
<- server decides and communicate back

vs just using the version are:
1. allow to add/tweak compression without version bump
2. allows to take into consideration ping time / system load

I think `MessageHeader` is even more flexible, since system load would change over time.

I am not completely sure why this can't be offloaded on the HTTP stack, I would assume perf might be an implementation detail, like curl + zstd or duckdb + zstd should perfrom about the same.

guillesd:
@carlopi I did some checks with gzip using curl ant it was terribly slow compared to duckdb + zstd. I did not know you could also do zstd with curl, may be worth a check... Although I'm not sure how it fits with the current model where:

- Executor threads do the serialization
- Async threads only do the sync requests as to free up executors and saturate the network

If we hand off compression to the async threads then we may not saturate the network. I rather do any CPU work on the executor thread.

PR #220, opened 2026-07-14
Add Client ID for attach

pdet:
Superseeds: https://github.com/duckdb/duckdb-quack/pull/183

This PR adds the possibility to manually pass a `client_id`. The client optionally passes a client_id string when attaching:
`ATTACH {server} AS remote (token 'asdf', client_id 'random_client_id')` The client_id is propagated to the server on the connection request, along with the token.

After successful authentication the server computes `client_id_hash = HMAC-SHA256(server_secret, client_id)`, where server_secret is a per-server, randomly generated private secret created at server startup. The hash is stored in the QuackConnection and exposed in quack_active_connections(), and it is also emitted on the server-side entries of the Quack logger, which makes tracking a specific client more explicit.

The hash is stable, as it is derived from the client_id only, so the same client_id will map to the same hash as long as the server is alive. And it is unforgeable as it is keyed with the server-private `server_secret`.


carlopi:
* `test/sql/quack_active_connections.test` doesn't parse
* maybe rename `server_secret` to `server_hmac_key` or equivalent, this is not a duckb `CREATE SECRET` nor actually confidential
* maybe minimal length to client_id ? token is >= 4, I guess same, to get some entropy in
* default `client_id` to something precomputed at `quack` load time? like we have some `whoami()` machinery that computes and stores `whoami_started_at`, same can be done for client_id. It can also arguably be init from env variable (allowing trivial operations on this). Idea is that evething will always get a `client_id`, but if not explicitly opted out OR if you don't use a setting like `SET quack_default_client_id = '';` or something
* code quality!! it's failing

carlopi (commented):
Thanks! Left some comments

carlopi (approved):
This looks solid, thanks!!

PR #222, opened 2026-07-16
Override httpfs settings in quack

Mytherin:
Currently quack inherits the httpfs timeout / retry settings which can cause issues both because the default settings are very restrictive, and because the user can override them. This PR makes it so that quack itself sets these instead of relying on the user settings.

CC @carlopi

### DIFF
--- a/src/include/quack_client.hpp
+++ b/src/include/quack_client.hpp
@@ -86,6 +86,8 @@ private:
 
 class HttpsQuackClient : public QuackClient {
 public:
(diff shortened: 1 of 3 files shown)


carlopi:
I would have assumed the correct position would have been before constructing the `PostRequestInfo` object, but I am not sure, just curious as to why it's there, maybe worth a comment?

Mytherin:
You're right, thanks for checking. I think this worked because of the client caching causing subsequent requests to use the new settings but that location was wrong for sure.

carlopi (approved):
Thanks!

PR #224, opened 2026-07-22
Add client acknowledgement after query completion

pdet:
This PR introduces an ACKNOWLEDGEMENT message that the client sends to the server once a query's results have been fully fetched, it is necessary for reconnect support. We implement a new `ACKNOWLEDGEMENT` message type, to which the server replies with a `SuccessResponse`. This is activated by a `quack_enable_reconnects` (which defaults to false). When that option is set to true, the scan sends an acknowledgment when the query completes.
  

Supersedes: https://github.com/duckdb/duckdb-quack/pull/181


guillesd (approved):
Looks great! Thanks!

PR #228, opened 2026-07-28
Server-Side Query Result Cache for Reconnects

pdet:
When a client dies mid-query today, the result it never received is simply lost, hence the query must be executed again from scratch. This PR adds the server half of `quack_enable_reconnects`. We implemented a query caching on the server-side, so the server retains each connection's last client query result. That means that a client that existed the connection mid query (e.g., a crash) can still recover that result.

This PR is only the server-side caching, so the final piece, where a reconnecting client can re-request the result for a `query_uuid` will come in a follow-up PR.

Each QuackConnection in the server's `active_connections` map holds a `result_cache` with its last client query's result. This allows us to hold multiple cached results one for each session. Retained chunks go into a buffer-managed ColumnDataCollection. 

The cache is cleared in a few situations:
* Client ACK - When a client sends an `ACKNOWLEDGEMENT` once it scanned the full result.
* Displacement - The connection's next client query clears the old cache before executing
* Cancel - Cancel drops the cache along with the result
* Disconnect
* TTL expiry 

The ttl expiry is traffic-driven, every inbound message will walk all connections and drop caches idle longer than `quack_result_ttl`. The idleness measumenet is compared against `last_served_at` which is refreshed on every served batch.

The following options are relevant:

| Setting                   | Type    | Default  | Description                                                                 |
|---------------------------|---------|----------|-----------------------------------------------------------------------------|
| `quack_enable_reconnects` | BOOLEAN | `false`  | Client: acknowledge fully-received results. Server: retain each connection's last result until acknowledged. |
| `quack_cache_max_rows`    | UBIGINT | `100000` | max rows retained per connection's replay cache. `0` = unlimited. |
| `quack_result_ttl`        | UBIGINT | `3600`   | Seconds an unacknowledged cached result may sit idle before any server expires it. `0` = never expire. |


guillesd:
Looked good to me but the clanker had some fair comments. Main one is (1), I really think we should avoid `SweepExpiredCaches` on every inbound message. I think (2) is also worth doing!

## 1. Gate the TTL sweep on the feature being on

`quack_server.cpp:372` runs `SweepExpiredCaches` on every inbound message, and it only bails when `quack_result_ttl == 0`. The default is 3600, so a server with reconnects off still pays, per message: a second acquire of the global `active_connections_mutex` (`GetConnection` already took it for the same message), a vector alloc, N refcount bumps, N `try_lock`s. All to find every cache null.

FETCH is our hottest message, so this lands on the read path everyone is already using. An early `if (!ServerCachingEnabled(db)) return;` makes it free.

> Note from Guille: maybe we should do a small benchmark to see how much this trips over performance in queries. Could also be worth to take a look at memory usage? If I see correctle, the cache calls `InitializeAppend(append_state)` in the constructor and never finalizes it, so the current append block stays pinned for the cache's whole life. That is one block per cached connection, not the whole collection, but it is unspillable and it scales with connection count.

## 2. ACK should carry a `query_uuid`

`AcknowledgementMessage` serializes nothing, so the handler at `quack_server.cpp:696` drops whatever is cached, no questions asked. If an ACK ever arrived while `cache->tail` was still live, the next FETCH returns an empty response and the client reads it as "done". Short result, no error.

I could not build a path where our client does this, it only ACKs after the scan is complete.

> Worth fixing anyway because every other message validates the uuid (FETCH gives "Result has been closed", CANCEL gives "cancel a different query"), and `ExpireCacheIfStale` deliberately marks `CANCELLED` rather than truncate. ACK is the one message that trusts the sender, and the one whose failure mode is silent.

## 3. Null-guard the other two setting readers

`ServerCachingEnabled` checks `!val.IsNull()`. `CacheMaxRows` and `ResultTtlMicros` don't, and all three throw away the `TryGetCurrentSetting` return. A NULL `Value` throws out of `GetValueInternal`.

Not reachable today, the settings are always registered and CI proves it including after a `RESET GLOBAL`. But `ResultTtlMicros` throws from inside `HandleMessage` before the connection is even resolved, so it would 500 every message. That is the same server-wide failure the `result_ttl_overflow.test` was written to prevent, so the intent is already there, just applied to one of the three. Two lines.

## 4. `quack_enable_reconnects` scope

Server reads it globally, client reads it session-first. So on a server `SET quack_enable_reconnects = true` does nothing and you need `SET GLOBAL`. The two settings added right next to it are both `SetScope::GLOBAL`.

Quiet footgun, and the tests already carry a comment explaining the trap, which is usually a sign someone will hit it.

## Nits

- `ServeBatch` checks the cap after the loop (`quack_result_cache.cpp:101`), so with `quack_cache_max_rows = 100` you still materialize a full chunk before dropping. Harmless, just reads odd for something called max rows.
- `cache` is a raw pointer that dangles right after `ClearResultCache()` at `:90` and `:104`. Correct as written, nothing touches it after, but fragile for whoever edits that loop next.
- `cached_rows` got inserted before `query_started_at` in `quack_active_connections()` rather than appended. Everything in repo selects by name so nothing breaks, but `SELECT *` consumers see a column move.
- The cache costs a full extra copy of every served row (`retained.Append` at `:98`). Expected, but relevant to the serialize-at-sink work, which is busy removing CDC copies from both paths.

## Testing

Sufficient, no new tests needed. `ServeBatch` and `HasMoreResults` are on the unconditional path for every PREPARE and FETCH, so the whole existing suite already covers the caching-off path. The four new files cover gating, displacement, internal traffic not displacing, ACK, the cap and cap=0, errors at prepare and mid-stream, TTL reap and survival and 0, both overflow modes, and multi-connection.

The only test I would add is one that can't exist yet: if point 2 gets fixed, assert that a mid-stream ACK errors instead of truncating.

pdet:
@guillesd I think I did all your changes plus some other optimizations.
The ACK now carries the query UUID, so an ACK can only clear the cache of the query it belongs to.  We have a counter to skip the TTL cleaning if nothing is cached, and I've added the NULL guards.

I've also changed the TTL protocol a bit, we not use a priority queue so we don't have to do the whole search space, and the expiration is off the request path.

LMK what you think.

src/include/quack_message.hpp
```diff
@@ -489,14 +489,23 @@ class AcknowledgementMessage : public QuackMessage {
...
+	hugeint_t QueryUUID() const {
+		return query_uuid;
+	}
+
 protected:
 	AcknowledgementMessage() : QuackMessage(TYPE) {
 	}
+
+private:
+	//! Query the client is acknowledging. {0,0} means unspecified (pre-uuid clients).
```
guillesd: This is not really handled in the server handler (which may be fine but I think the comment misleads a bit)

src/quack_server.cpp
```diff
@@ -657,12 +739,28 @@ unique_ptr<QuackMessage> QuackServer::HandleMessageInternal(DatabaseInstance &db
...
 		connection.query_state = QuackQueryState::CANCELLED;
 		connection.duckdb_query_result.reset();
+		connection.ClearResultCache();
 		return make_uniq<SuccessResponse>();
 	}
 	case MessageType::ACKNOWLEDGEMENT: {
+		auto &acknowledgement_message = received_message.Cast<AcknowledgementMessage>();
+		auto &connection = *connection_p;
+		std::unique_lock<std::mutex> lock(connection.lock);
+		auto &cache = connection.result_cache;
```
guillesd, line 763: Back to the message comment, we could do this here:
```cpp
auto ack_uuid = acknowledgement_message.QueryUUID();
// A pre-uuid client cannot name the query it is acknowledging, so fall back to clearing
// whatever is cached rather than failing its ack and leaking the cache until the TTL.
if (ack_uuid != hugeint_t {0, 0} && cache->query_uuid != ack_uuid) {
      return make_uniq<ErrorResponse>("Attempted to acknowledge a different query with id '%s' instead of '%s'",
                                      ack_uuid, cache->query_uuid);
}
connection.ClearResultCache();
```

But it may be fine also if we don't.. not a strong opinion

src/quack_server.cpp
```diff
@@ -150,7 +152,25 @@ QuackServer::QuackServer(ClientContext &context_p, const QuackUri &uri_p, const
...
+	}
+	// the thread owns only the doomed caches (plus the db keeping their buffers valid), never the server
+	std::thread([db = std::move(db), doomed = std::move(doomed)] {}).detach();
+}
+
 vector<QuackConnectionSnapshot> QuackServer::GetActiveConnectionSnap() {
+	auto db = db_ptr.lock();
+	if (db) {
+		// reap first so an expired-but-unswept cache is never reported as live
+		SweepExpiredCaches(*db);
```
guillesd: Clanker says this leads to undefined behaviour:
> Because we reached the sweep from inside a request that already holds connection C's lock, SweepExpiredCaches pops C's own expiry slot, GetConnection hands back C, and try_lock is then called on C's mutex — the same mutex this same thread is already holding. Self-locking a std::mutex is undefined behaviour

src/quack_extension.cpp
```diff
@@ -212,8 +212,17 @@ static void LoadInternal(ExtensionLoader &loader) {
 	                          LogicalType::VARCHAR, Value(default_client_id), nullptr, SetScope::GLOBAL);
 
 	config.AddExtensionOption("quack_enable_reconnects",
-	                          "Send an acknowledgement to the server after a query completes", LogicalType::BOOLEAN,
-	                          Value::BOOLEAN(false));
+	                          "Enable reconnect support (clients acknowledge results, the server caches the last "
```
guillesd, line 215: any reason why this is not `SetScope::Global`? the other two added options are!

guillesd (commented):
Just a couple of things! Also, there are some merge conflicts.

src/quack_extension.cpp
pdet: I think it makes sense, since on the server side it needs to be global, but on the client side it is dependent on the specific session.

guillesd (approved):
Super

PR #229, opened 2026-07-28
Feat/fetch read ahead v2

guillesd:
Builds on #208. That PR made the client fetch ahead, but the server side was still a single lazily driven cursor behind a mutex: one thread pulling chunks out of a `StreamingQueryResult`, serializing them on the reply thread, and handing out "the next batch" to whoever asked. That design has three problems: production is serial no matter how parallel the query is, every FETCH pays serialization latency on the critical path, and "give me the next batch" is unsafe to retry (a lost response means a lost batch).

This PR replaces that with a parallel, pre-serialized, retry-safe read path. The query now runs into a result collector that produces wire-ready batches on the executor threads, and FETCH becomes an addressed, idempotent request: the client asks for batch N and the server can answer that exact question any number of times.

> A follow up PR will move the send data path into the same machinery

## The core: claim buffer and rebalancer

Two components carry the design, and both are deliberately generic because the send path will reuse them:

**`QuackClaimBuffer`** is the delivery primitive: a buffer of densely numbered batches (1, 2, 3, ...) where each consumer claims the next index and waits for exactly that batch. Producers can push out of order; consumers still see a monotone sequence per thread. It also carries the safety features: push dedupe (a retried delivery of the same index is dropped), byte-capacity backpressure into the producer, head of stream admission (the batch the consumer needs next is always admitted so a full buffer of later batches cannot starve it), and `Finish(expected_total)` which errors if the stream closed short instead of silently truncating.  The claiming model is borrowed from core's multi-file read-ahead (consumers claim the next ticket and wait for exactly that one), and waiting reuses its `ReadAheadJobCompletion` directly, so a scan thread whose batch is not there yet parks and returns `BLOCKED` to the executor, and publishing that batch wakes exactly that task. The buffer itself is hand-rolled: core's machinery is a prefetch task runner, not an addressable batch buffer. Converging the two in core someday would be nice. The same template backs the client fetch buffer and the server fetch stream, and later the insert stream.

**`QuackRebalancerCore/Sink`** solves the shape mismatch between what the executor produces and what we want on the wire. Executor batches are sparse and uneven (a filtered scan can produce 1kb fragments); wire messages want to be dense and evenly sized. The rebalancer accumulates fragments per thread up to `quack_target_batch_bytes` (default 32MB), maps them onto dense indices while preserving source order when required, and emits them out of order (receivers reassemble by index). The key property is that fragments are accumulated directly in their final serialized form, so cutting a batch costs no extra copy or serialization pass. On this PR the rebalancer feeds the server fetch collector; on the next one the same sink feeds the client INSERT.

A note on that last point, since the obvious alternative is buffering chunks in a `ColumnDataCollection` and serializing at send time (as I suggested in #191). Serializing into a `MemoryStream` as chunks arrive wins on three counts: the data is copied exactly once (chunk to wire bytes, on the producer thread, instead of chunk to CDC to wire bytes with the serialization pass landing on whichever thread cuts the batch), batches get cut on their actual serialized size rather than a columnar in-memory estimate, and the server can hand the exact same buffer to httplib for the response and retain it for retries with zero further work. The trade-off is losing the buffer manager backing a CDC would give us; the claim buffer's byte cap and the rebalancer's memory budget bound it instead.

## Life a chunk in the fetch path

Diagram by Claude et al.

```txt
             SERVER               |               CLIENT
                                  |
  query starts on a background <--+---  PREPARE(sql)
  thread with the fetch           |
  collector plugged in            |
      |                           |
      | DataChunks (parallel      |
      | executor threads)         |
      v                           |
  fragment builder: chunks        |
  serialize straight into a       |
  FETCH_RESPONSE payload as       |
  they arrive                     |
      |                           |
      | seal at the byte target   |
      v                           |
  rebalancer: sparse executor     |
  batches become dense 1,2,3,...  |
  stamped into the serialized     |
  payload (8-byte patch)          |
      |                           |
      v                           |
  claim buffer (byte-capped,      |
  blocks the executor if the      |
  client falls behind)            |
      |                           |
      | leading batches inlined --+-->  scan starts on the
      | into the PREPARE response |     inline chunks
      |                           |         |
      |                           |         v
      |                           |     QuackFetcher keeps up
      |                           |     to K FETCHes in flight
      |                           |     on the async pool
      |                           |         |
      | <-------------------------+---  FETCH(batch=N, ack=M)
      v                           |
  pop batch N, patch the          |
  client-visible index, retain    |
  the payload until acked         |
      |                           |
      | raw bytes, zero copy,   --+-->  decode off the pipeline
        nothing serialized on     |     threads, publish into
        the reply thread          |     the claim buffer
                                  |         |
                                  |         v
                                  |     scan threads claim the
                                  |     next index and wait for
                                  |     exactly that batch
                                  |         |
                                  |         v
                                  |     DataChunks flow into
                                  |     the local pipeline
```

Note the reply thread does no serialization work at all: it pops a finished payload, patches 8 bytes, and hands the buffer to httplib via a content provider (no copy either).

## Retry safety

Every FETCH names its batch and carries an ack watermark ("I have everything up to M"). The server retains served payloads until acked, so an HTTP-level retry of a lost response re-serves the same batch instead of popping the next one and losing it. On top of that, both ends validate the total batch count at end of stream, so a lost batch fails loudly. Duplicate deliveries are dropped by the buffer's dedupe.

> Small reads pay a PREPARE latency cost because the response has to wait for the first batch to seal. The fix (sealing the first batch early / opportunistically) is designed and queued as a follow-up PR.

## Follow-ups

- Send path on the same rebalancer + claim buffer (next PR in the stack).
- Early seal for PREPARE latency on large batches.
- Upstream reserved/backpatchable field support in DuckDB's serializer, which would replace the hand-mirrored payload writer and its debug pin entirely (see annex).
- Evicting the retained tail at end of stream. The last ~depth payloads currently stay retained until the next PREPARE or disconnect because nothing acks them. Plan is a best-effort FINALIZE from the client at scan end, which also tears down abandoned scans; needs a quick alignment with #224's ACKNOWLEDGEMENT since they overlap.

## Annex: design notes

I think you will see some interesting/questionable decissions and I wanted to get ahead by providing some explanations. This doesn't make the decisions unquestionable but hopefully it saves a conversation round trip (as we like to optimize round trips here).

**Why batches are stamped after serialization, and the weird serializer that enables it.** The rebalancer only knows a batch's dense index at release time, after the batch is fully serialized, and the FETCH handler renumbers it a second time at serve time (the client-visible index subtracts whatever PREPARE consumed inline). Patching a normal serialized integer is impossible because the encoding is variable width, so the index travels as `QuackBatchIndexField`: a fixed 8-byte field, serialized last, overwritten in place. The chunk count has the same problem in reverse (written before the chunks, known after), solved by zero-padding its LEB128 to a fixed 5 bytes; padded LEB128 decodes to the same value, so standard readers are unaffected.

**`QuackChunkPayloadWriter` and the debug round-trip pin.** Streaming chunks into a payload means hand-writing an incremental encoder that must stay byte-identical to the generated codec the receiver uses. That mirror is convention, not compiler-checked, so `Seal()` in debug builds decodes its own bytes with the generated codec, re-encodes, and requires equality (count region compared by value because of the padding). Any drift between the two encoders fails the first debug test instead of surfacing as a wire error weeks later. To be clear, this whole construction (the hand-mirrored writer, the placeholder tricks, the pin that guards them) exists because DuckDB's serializer has no notion of reserved or backpatchable fields. The proper fix is to add that support in core: let a serializer reserve a fixed-width slot and patch it after the fact, and the writer collapses into the generated code path with no drift to guard against.

**`FetchResponseMessage` codec stays generated.** The fixed-width field needs custom encode/decode logic, but instead of hand-maintaining the codec we use the generator's own hooks (`serialize_property`, `deserialize_skip_assign`, `methods`), so `quack_message.json` remains the wire truth and `generate_serialization.py` is safe to run.

**PREPARE still inlines leading rows.** Small results (catalog enumeration in particular) complete in one round trip. Since batches are stored pre-serialized, the inline drain is the one place that decodes them back into chunks, and FETCH indices are remapped by the consumed count. The old `quack_fetch_batch_rows` setting only governed this drain anymore, so it is renamed to `quack_prepare_inline_rows`.

**Head of stream admission.** Emission is parallel and out of order, but the server only ever pops the exact index the client asked for. Without an exception, a full buffer of later batches can block the one batch the client is waiting on (deadlock-shaped stall, reachable with a single-threaded client). The claim buffer therefore always admits the head of the stream, overshooting the cap by at most one batch. `quack_debug_emit_delay_ms` exists to make tests hit this shape reliably.

**Coexistence with #224.** The ACKNOWLEDGEMENT message and reconnect tests are preserved. Note there are now two unrelated "ack" concepts: #224's query-completion ACK (reconnect support, opt-in) and this PR's per-FETCH `ack_index` watermark (served-payload GC). This may create some confusion, so we should consider a clearer naming convention for either of the two.

## Important EDIT --> we should add backpatching in core first

I think that to make this PR clean we should make backpatching possible via extending the binary serializer in core and the `WriteStream`. Propposal:

### For `WriteStream`

```cpp
class WriteStream {
public:
      virtual void WriteData(const_data_ptr_t buffer, idx_t write_size) = 0;
      // Backpatching: streams that keep written bytes reachable can patch them in place.
      virtual bool SupportsBackpatching() const {
              return false;
      }
      virtual idx_t GetStreamPosition() const;             // throws if unsupported
      virtual void PatchData(idx_t offset, const_data_ptr_t buffer, idx_t size); // throws if unsupported
};
```
Which then `MemoryStream` implements.

### For the `Serializer`

```cpp
class Serializer {
public:
      // A handle to a fixed-width slot: plain offset + width, so the payload owner
      // can re-patch the raw retained bytes later without a live serializer.
      struct ReservedSlot {
              idx_t offset = DConstants::INVALID_INDEX;
              idx_t width = 0;
      };

      // Writes the field header plus a padded-varint placeholder of 'width' bytes.
      virtual ReservedSlot WriteReservedProperty(field_id_t field_id, const char *tag, idx_t width);
      virtual void PatchReserved(const ReservedSlot &slot, uint64_t value);

      // A list whose count is unknown until the elements have been written.
      virtual ReservedSlot BeginDeferredList(field_id_t field_id, const char *tag, idx_t count_width);
      template <class T>
      void WriteDeferredListElement(const T &value); // same element encoding as WriteList's body
      virtual void EndDeferredList(const ReservedSlot &slot, idx_t count);
};
```

And then implemented by the `BinarySerializer`:

```cpp
Serializer::ReservedSlot BinarySerializer::WriteReservedProperty(field_id_t field_id, const char *tag, idx_t width) {
      OnPropertyBegin(field_id, tag);
      ReservedSlot slot {stream.GetStreamPosition(), width};
      data_t padded[MAX_RESERVED_WIDTH];
      EncodePaddedLEB128(0, padded, width); // redundant continuation bits, ProtoZero-style
      stream.WriteData(padded, width);
      OnPropertyEnd();
      return slot;
}

void BinarySerializer::PatchReserved(const ReservedSlot &slot, uint64_t value) {
      D_ASSERT(value < (1ULL << (7 * slot.width))); // must fit the reserved width
      data_t padded[MAX_RESERVED_WIDTH];
      EncodePaddedLEB128(value, padded, slot.width);
      stream.PatchData(slot.offset, padded, slot.width);
}
```

Which is what we basically rolled out in Quack now (and the reason we have basically two potentially divergent serializers)

### DIFF
--- a/src/include/quack_message.hpp
+++ b/src/include/quack_message.hpp
@@ -2,11 +2,14 @@
 
 #include "duckdb/common/serializer/binary_serializer.hpp"
 #include "duckdb/common/serializer/memory_stream.hpp"
+#include "duckdb/common/types/data_chunk.hpp"
 #include "duckdb/parser/parsed_data/drop_info.hpp"
 #include "duckdb/common/types/uuid.hpp"
 #include "duckdb/common/error_data.hpp"
 
 namespace duckdb {
+
+class ClientContext;
 
 //! Quack wire-protocol version. Client and server agree on it during the connection handshake.
 static constexpr idx_t QUACK_VERSION = 2;
@@ -54,6 +57,31 @@ private:
 
 string MessageTypeToString(MessageType type);
 
+//! Fixed-width (8-byte LE) batch-index wire field, serialized LAST so an already-encoded payload can
+//! be stamped in place. 0 = invalid/absent; PLACEHOLDER = never stamped (receivers reject it).
+struct QuackBatchIndexField {
+	static constexpr uint64_t PLACEHOLDER = 0xF1DCBA99C0FFEE42ULL;
+	static constexpr uint64_t INVALID = 0;
+
+	static void Encode(uint64_t value, data_ptr_t out) {
+		for (idx_t i = 0; i < 8; i++) {
+			out[i] = static_cast<data_t>((value >> (i * 8)) & 0xFF);
+		}
+	}
+	static uint64_t Decode(const_data_ptr_t in) {
+		uint64_t value = 0;
+		for (idx_t i = 0; i < 8; i++) {
+			value |= static_cast<uint64_t>(in[i]) << (i * 8);
+		}
+		return value;
+	}
+	//! Stamp the dense index into an encoded payload.
+	static void Patch(data_ptr_t payload, idx_t payload_size, idx_t offset, idx_t batch_index) {
+		D_ASSERT(offset + 8 <= payload_size);
+		Encode(batch_index, payload + offset);
+	}
+};
+
 struct MessageHeader {
 	MessageHeader(MessageType type_p, string connection_id_p)
 	    : type(type_p), connection_id(std::move(connection_id_p)) {
@@ -69,12 +97,19 @@ struct MessageHeader {
 
 class QuackMessage {
 public:
-	void ToMemoryStream(MemoryStream &write_stream) const;
+	virtual void ToMemoryStream(MemoryStream &write_stream) const;
 	static unique_ptr<QuackMessage> FromMemoryStream(MemoryStream &read_stream);
+
+	//! Pre-serialized payload carried instead of a serializable body (see QuackRawPayloadResponse);
+	//! the HTTP layer sends these bytes directly. Null for normal messages.
+	virtual optional_ptr<MemoryStream> RawPayload() const {
+		return nullptr;
+	}
 
 	template <class TARGET>
 	TARGET &Cast() {
-		if (header.type != TARGET::TYPE) {
+		// a raw-payload carrier shares the payload's type tag but not its layout
+		if (header.type != TARGET::TYPE || RawPayload()) {
 			throw InternalException("Failed to cast message to type - message type mismatch");
 		}
 		return reinterpret_cast<TARGET &>(*this);
@@ -82,7 +117,7 @@ public:
 
 	template <class TARGET>
 	const TARGET &Cast() const {
-		if (header.type != TARGET::TYPE) {
+		if (header.type != TARGET::TYPE || RawPayload()) {
 			throw InternalException("Failed to cast message to type - message type mismatch");
 		}
 		return reinterpret_cast<const TARGET &>(*this);
@@ -282,8 +317,8 @@ class FetchRequestMessage : public QuackMessage {
 public:
 	static constexpr MessageType TYPE = MessageType::FETCH_REQUEST;
 
-	explicit FetchRequestMessage(string connection_id_p, hugeint_t uuid)
-	    : QuackMessage(TYPE, std::move(connection_id_p)), uuid(uuid) {
+	FetchRequestMessage(string connection_id_p, hugeint_t uuid, idx_t batch_index, idx_t ack_index)
+	    : QuackMessage(TYPE, std::move(connection_id_p)), uuid(uuid), batch_index(batch_index), ack_index(ack_index) {
 	}
 
 protected:
@@ -295,8 +330,15 @@ public:
 	static unique_ptr<FetchRequestMessage> Deserialize(Deserializer &deserializer);
 
 	hugeint_t uuid;
+	//! Client-visible dense batch index (from 1) this request claims. Every request names its batch,
+	//! so a transport-level retry re-asks for the SAME batch instead of popping the next one.
+	idx_t batch_index = 0;
+	//! Highest index with all of 1..ack_index received; the server may drop retained batches <= it.
+	idx_t ack_index = 0;
 };
 
+// One dense batch of a query result (server -> client). Data responses are pre-serialized by
+// FetchResponsePayloadWriter; this class serializes only the empty terminal FINISHED response.
 class FetchResponseMessage : public QuackMessage {
 public:
 	static constexpr MessageType TYPE = MessageType::FETCH_RESPONSE;
@@ -310,6 +352,11 @@ public:
 	void Serialize(Serializer &serializer) const override;
 	static unique_ptr<FetchResponseMessage> Deserialize(Deserializer &deserializer);
 
+	//! batch_index encoded as the fixed-width QuackBatchIndexField wire string (generator hook).
+	string EncodeBatchIndexFixed() const;
+	//! Decode + validate the wire string back into batch_index; rejects unstamped payloads.
+	void ApplyBatchIndexFixed(const string &bytes);
+
 	vector<unique_ptr<DataChunkWrapper>> &MutableResults() {
 		return results;
 	}
@@ -318,9 +365,18 @@ public:
 		return batch_index;
 	}
 
+	//! Client-visible total batch count; only set on the terminal FINISHED response.
+	void SetTotalBatches(optional_idx total_batches_p) {
+		total_batches = total_batches_p;
+	}
+	optional_idx TotalBatches() const {
+		return total_batches;
+	}
+
 private:
 	vector<unique_ptr<DataChunkWrapper>> results;
 	optional_idx batch_index;
+	optional_idx total_batches;
 };
 
 // Streams one or more DataChunks of insert data to the server, keyed by (connection_id, query_uuid).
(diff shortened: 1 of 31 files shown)


guillesd:
cc @Mytherin

src/include/quack_message.hpp
```diff
@@ -295,8 +330,15 @@ class FetchRequestMessage : public QuackMessage {
 	static unique_ptr<FetchRequestMessage> Deserialize(Deserializer &deserializer);
 
 	hugeint_t uuid;
+	//! Client-visible dense batch index (from 1) this request claims. Every request names its batch,
+	//! so a transport-level retry re-asks for the SAME batch instead of popping the next one.
+	idx_t batch_index = 0;
```
Mytherin, line 348: Should this be optional? If not, how do we know we're done?

src/include/quack_message.hpp
```diff
@@ -318,9 +365,18 @@ class FetchResponseMessage : public QuackMessage {
 		return batch_index;
 	}
 
+	//! Client-visible total batch count; only set on the terminal FINISHED response.
```
Mytherin: If we set total_batches whenever it is known server side we can avoid a final (pointless) request to get this (i.e. if we have 3 total batches, we can already send back `total_batches = 3` as soon as we know it, preventing a request for a non-existing batch 4).

src/include/quack_rebalancer_core.hpp
```diff
@@ -0,0 +1,160 @@
...
+	virtual unique_ptr<QuackPreparedBatch> Seal(ClientContext &context) = 0;
+};
+
+//! Compacts sub-half-vector chunks into staged ~STANDARD_VECTOR_SIZE chunks so sparse sources
+//! (filtered scans) don't pay per-chunk framing overhead; larger chunks pass straight through.
+class QuackChunkStager {
+public:
+	template <class FLUSH>
+	void Append(DataChunk &chunk, FLUSH &&flush) {
+		if (chunk.size() * 2 >= STANDARD_VECTOR_SIZE) {
```
Mytherin, line 36: Why `2 *` here?

src/quack_fetch_collector.cpp
```diff
@@ -0,0 +1,199 @@
...
+
+	unique_ptr<QueryResult> GetResult(GlobalSinkState &state) const override {
+		// the data left through the stream's claim buffer; the statement's own result is empty
+		auto &gstate = state.Cast<QuackRebalancerGlobalState>();
+		auto context = gstate.client_context.lock();
+		if (!context) {
+			throw InternalException("client context expired in QuackFetchCollector::GetResult");
+		}
+		auto collection = make_uniq<ColumnDataCollection>(Allocator::DefaultAllocator(), types);
+		return make_uniq<MaterializedQueryResult>(statement_type, properties, IdentifiersToStrings(names),
```
Mytherin: Should this not be a different result type (e.g. `SERIALIZED_MESSAGE_RESULT`)?

Mytherin: You can look at `ArrowQueryResult` for inspiration

src/quack_server.cpp
```diff
@@ -449,94 +534,163 @@ unique_ptr<QuackMessage> QuackServer::HandleMessageInternal(DatabaseInstance &db
...
-				return response;
+			// Swap-install atomically: a concurrent PREPARE may have installed a producer since the
+			// abort above, and assigning over a joinable std::thread would terminate the process.
+			lock_guard<mutex> guard(connection.fetch.lock);
+			superseded.stream = std::move(connection.fetch.stream);
+			superseded.thread = std::move(connection.fetch.thread);
+			connection.fetch.stream = stream;
+			connection.fetch.uuid = prepare_request_message.QueryUUID();
+			connection.fetch.abort_error = ErrorData();
+			connection.fetch.thread = std::thread(RunFetchQuery, std::ref(connection), stream, effective_sql);
```
Mytherin, line 772: Can you have a look at this thread usage here and why it's required?

Mytherin: If we have to use a thread here (which I think we will have to eventually, but not sure if it's part of this PR - maybe not) - we should use a thread pool or use tasks that are part of the databases' thread pool

Mytherin (commented):
Thanks for the PR! Looks great - some comments

src/include/quack_message.hpp
guillesd: This is for the request actually, and the client never knows that there's no more data left, so it will always claim a batch. In the `FetchResponseMessage` we do have `total_batches` as `optional_idx`

src/include/quack_message.hpp
guillesd: This is absolutely true. We are sending a last response with no data and only the total amount of batches. The easy fix is to do the same backpatching for the `total_batches`. We reserve a slot for it in every message and stamp it when the query has finalized on every fetch response. Then the client knows not to emit tickets that are passed that value. I wonder if for small queries prefetching will maybe send more requests than it needs because the responses with the `total_batches` stamped are late to the party.

If you think this is a good idea I can rework it a bit (it shouldn't be much)

src/include/quack_rebalancer_core.hpp
guillesd: This is made up heuristics. The idea is the following: we want to avoid sending fragments of chunks (50, 60, 70 row chunks), but we accept half a chunk and bigger because it is preferable to serialize directly than to do a `memcpy`. To be honest this heuristic is made by Claude who thinks the following:

> Half is also the natural cut for a merging heuristic: combining sub-half chunks can reduce chunk count by ≥2×, while merging chunks above half can't.

But is definitely not based on experimentation and we could remove the 2.

src/quack_fetch_collector.cpp
guillesd: Regarding this we can do one of two:
1) We merge this PR and I make later a separate PR that adds `SERIALIZED_MESSAGE_RESULT` to the `QueryResultType` enum (and we can decide if `class QuackFetchQueryResult : public QueryResult` should be in that PR or if we just keep it in Quack.

2) We don't merge this PR and I work on the core PR first.

If we choose 1) I can add some tasks to the Quack tracker for 2.0

src/quack_server.cpp
guillesd: This is the wrong machinery.. like it works now but it parks a thread populating the fetch stream. With multiple concurrent queries this means many threads parked. I'm scoping a better alternative. If it is not a super big change then I may include it in this PR. If it is large maybe we can keep this as is and I can do a follow up.

guillesd:
I'm tracking the follow up prs (including the ones needed for core) here https://github.com/duckdblabs/duckdb-internal/issues/9844

src/quack_server.cpp
guillesd: Okey.. I took a look at this. With the current model, where we always have a `result->Fetch()` we need something to drive fetching results that is non-blocking for the pipeline execution and the actual send of fetch responses. Ergo we have this "driver" threads (one per connection) that are there basically to fill up the `QuackFetchStream` buffer.

With the changes that I made in `de6dc10` i make use of the `InterruptState` machinery so that the sink can emit a `BLOCKED` status, which at least means that when the buffer is full the producers yield back to the scheduler.

However yes, this may be a problem if there are multiple concurrent connections and queries in flight. We may need to redesign this a bit: either by having a elastic pool of driver threads or something different? I think this can be a follow up PR.

Mytherin:
Thanks!

PR #231, opened 2026-07-29
Fix ATTACH when remote tables have computed DEFAULTs (#132)

### LINKED ISSUE #132: ATTACH fails if there is a table with a field which has a computed DEFAULT value
This is my example:

`server.sql`
```
LOAD quack;

CREATE TABLE t (x INTEGER NOT NULL default 1);
INSERT INTO t DEFAULT VALUES;
FROM t;

CALL quack_serve('quack:localhost', token => 'tok1234');
```
`client.sql`
```
LOAD quack;

FROM quack_query('quack:localhost', ' FROM t', token => 'tok1234');

ATTACH 'quack:localhost' AS r (TYPE quack, TOKEN 'tok1234');
FROM r.t;
FROM r.query('from t')
```

I run the server with `duckdb -init server.sql`

then `duckdb < client.sql` gives me
```
❯ duckdb  < client.sql
┌───────┐
│   x   │
│ int32 │
├───────┤
│     1 │
└───────┘
┌───────┐
│   x   │
│ int32 │
├───────┤
│     1 │
└───────┘
┌───────┐
│   x   │
│ int32 │
├───────┤
│     1 │
└───────┘
```

now when I change the `server.sql`
and do this instead 
`CREATE TABLE t (x INTEGER NOT NULL default 1+0);  -- changed from 1 to 1+0 `

now only the first query works, but not the `ATTACH`
```
❯ duckdb  < client.sql
┌───────┐
│   x   │
│ int32 │
├───────┤
│     1 │
└───────┘
Binder Error:
Catalog "r" does not exist!
...
```

asubbarao:
## What happened

On **client ATTACH**, quack loads the remote catalog by:

1. Running `duckdb_tables()` on the server (gets each table’s `CREATE TABLE` SQL)
2. Parsing that SQL on the **client**
3. Calling `BindCreateTableInfo` — which **rebinds every column DEFAULT on the client**

**Computed** DEFAULTs (`1+0`, `current_timestamp`, `now()`, …) fail that client-side bind.  
The attach alias is never registered, so the only error the user sees is the misleading:

```text
Binder Error: Catalog "r" does not exist!
```

**Constant** DEFAULTs (`DEFAULT 1`) already worked.  
`quack_query` still works in the broken case (it never rebuilds a full client catalog), which made this easy to misdiagnose.

Postgres/Snowflake attaches do not rebind server DEFAULT expressions into the client catalog the same way — they expose names/types for planning. Quack should behave like that for catalog load.

Closes / fixes #132.

## How to repro

**Server** (`duckdb -init server.sql`):

```sql
INSTALL quack FROM core_nightly; LOAD quack;
-- works with ATTACH:
-- CREATE TABLE t (x INTEGER NOT NULL DEFAULT 1);
-- breaks ATTACH:
CREATE TABLE t (x INTEGER NOT NULL DEFAULT 1 + 0);
INSERT INTO t DEFAULT VALUES;
CALL quack_serve('quack:localhost', token => 'tok1234');
```

**Client**:

```sql
INSTALL quack FROM core_nightly; LOAD quack;
-- still works:
FROM quack_query('quack:localhost', 'FROM t', token => 'tok1234');
-- fails with Catalog "r" does not exist:
ATTACH 'quack:localhost' AS r (TYPE quack, TOKEN 'tok1234');
FROM r.t;
```

Same pattern with `DEFAULT current_timestamp` (e.g. live lookdown tables).

## Fix

In `QuackTableSet` (client catalog load), **strip column DEFAULTs after parse and before `BindCreateTableInfo`**.

- Client catalog only needs **names/types** for planning (same idea as other remote attaches).
- Real DEFAULTs stay on the **server**.
- `INSERT DEFAULT VALUES` still executes **remotely**.
- If bind still fails for another reason, the error now includes catalog name, schema, and the CREATE SQL (instead of only “catalog does not exist”).

## Test

`test/sql/attach_computed_default.test` — ATTACH with:

- constant `DEFAULT 1`
- computed `DEFAULT 1 + 0`
- `DEFAULT current_timestamp`

then scan + remote `INSERT DEFAULT VALUES`.

## Local confirmation

On a live quack server with lookdown tables using `DEFAULT current_timestamp` / `now()`:

- ATTACH failed with `Catalog "q" does not exist`
- Dropping those DEFAULTs unblocked ATTACH (workaround)
- This fix is the proper client-side solution so DEFAULTs can stay on the server

### DIFF
--- a/src/storage/quack_table.cpp
+++ b/src/storage/quack_table.cpp
@@ -3,6 +3,7 @@
 #include "quack_scan.hpp"
 #include "duckdb/parser/parser.hpp"
 #include "duckdb/parser/statement/create_statement.hpp"
+#include "duckdb/parser/parsed_data/create_table_info.hpp"
 #include "duckdb/planner/parsed_data/bound_create_table_info.hpp"
 #include "duckdb/parser/parsed_data/create_view_info.hpp"
 #include "storage/quack_view.hpp"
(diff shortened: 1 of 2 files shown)


asubbarao:
Superseded by #264, which targets `main` (this branch targeted `v1.5-variegata`) and fixes the approach.

The `SetDefaultValue(nullptr)` strip here is wrong: the attached table forgets the DEFAULT entirely, so `INSERT … DEFAULT VALUES` fails the NOT NULL constraint — it breaks this PR's own test at line 52, contradicting the claim in the description that remote `DEFAULT VALUES` still executes.

#264 uses `AlterBindMode::SKIP_BINDING` instead, so the parsed DEFAULT stays on the column and binds on first use once the catalog exists. Verified on quack `main` + duckdb `f1c54da7`: 78 tests, 0 failures.

PR #237, opened 2026-07-31
Clear socket options in HttpQuackServer unsetting SO_REUSEPORT

frankier:
The default httplib socket options has SO_REUSEPORT. This makes it very easy to bind two separate Quack servers to the same port which is not desirable/useful and quickly results in breakage in practice.

You can reproduce the problem by opening the `duckdb` CLI in two terminals and then running e.g.:

```
CALL quack_serve(
 'quack:localhost:9494',
 token = 'token'
 );
```

in both windows. You will notice both succeed unlike running it twice in a single DuckDB session. Trying to attach after this causes various problems, I'm guessing a TCP session is opened per-message and so things get interleaved and cause problems. At any rate, this is probably not what is intended so this patch just clears the default socket options.


frankier:
I think this one should be backported to `v1.5-variegata` too.

frankier:
Not sure what the problem was there. The tests pass locally.

PR #242, opened 2026-08-04
Reclaim sessions on socket close + give each by_name scan its own session

h4nsmuller:
Three related fixes to server-side session handling, motivated by "random behaviour when quack nodes read many S3 files concurrently."

### 1. Reclaim server-side sessions when their client sockets close (original PR)
`active_connections` sessions were only freed on an explicit `DISCONNECT`; a client that just drops its socket leaked its session (and the DuckDB connection it pins). Adds a reaper that reclaims sessions whose sockets are gone (`socket_refs == 0`) once idle past a grace, plus per-socket ref-counting (`ClaimSessionForCurrentSocket`/`ReleaseCurrentSocketSessions`). Careful with long-running requests: the serving socket holds a ref for the request's whole duration, and reconnects re-claim within the grace.

### 2. Reaper TOCTOU fix
`HandleMessage` looked a session up with `GetConnection` (which releases the map lock) and only *then* bumped `socket_refs` — so the reaper could observe `socket_refs == 0` in the gap, erase the session, and orphan the in-flight request (next request → "Invalid connection id"). New `GetConnectionForCurrentSocket` looks up **and** claims the ref while still holding `active_connections_mutex`, closing the window.

### 3. `quack_query_by_name`: own session per scan
The attached-catalog path reused the catalog's **single** `client_connection` (one `connection_id`) for every scan. The server keeps exactly **one** `duckdb_query_result` per `connection_id`, so two concurrent scans to the same attached database clobber each other — the second's `PREPARE` discards the first's in-flight result and resets `query_uuid`, and the first's next `FETCH` fails the uuid check (**"Result has been closed"**) or returns short/partial rows. A big many-file read holds its result open longer → wider clobber window → **"random behaviour reading many files at once."**

`QuackCatalog::CreateClientConnection()` now opens a fresh `connection_id` per scan (mirroring the direct `quack_query` path), which the socket-close reaper (fix #1) reclaims — so it fixes the correctness bug without reintroducing the session leak. The single catalog connection is still used for metadata/DDL.

### DIFF
--- a/src/include/quack_server.hpp
+++ b/src/include/quack_server.hpp
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <condition_variable>
 #include <thread>
 
 #include "duckdb/common/atomic.hpp"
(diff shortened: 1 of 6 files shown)


Mytherin:
This has been superseded by the heartbeat

matteo-emob:
Hi @Mytherin, does the heartbeat solve also point "3. quack_query_by_name: own session per scan" which could result in failing the uuid check, reporting error "Result has been closed"?

PR #245, opened 2026-08-12
Client/Server Heartbeat

pdet:
This PR introduces a heartbeat to the quack protocol. The issue was that dead clients (e.g., if not properly disconnected) wouldn't be cleared from the server.
This becomes an issue during reconnects, as we can't validate a client's uniqueness.
So we now add a lease to the connections. The lease is renewed by either a valid message or by a background heartbeat from the client. The server expires sessions when a lease lapses.

The heartbeat timeout can be set via `heartbeat_timeout`, during an `ATTACH`, and it defaults to 30 seconds. Connections will then run a background heartbeat thread that fires at ~1/3 of the timeout with a jitter of +/- 20%.

The server implements a background reaper that checks the current sessions every 100ms and expires lapsed ones.


carlopi (commented):
> Do you need an explicit reaper? Could that not be scheduled only on re-connects attempts?

That would be my comment, reducing (or moving) some complexity, plus handling freeze in the network for everyone nicely.

PR #248, opened 2026-08-15
Allow attaching catalogs with foreign keys

skuirrels:
Quack binds remote table DDL while building the attached catalog. Foreign-key binding looks for referenced tables in the client catalog, so `ATTACH` fails.

Skip foreign keys when binding remote table metadata. They remain server-side constraints and are still enforced on remote inserts.

Adds coverage for attach, reads, and valid and invalid remote inserts.

Checks:
- `make format-check`
- `make test` (34 tests, 635 assertions)
- debug regression test (13 assertions)


skuirrels:
Hi, any comments on this one as it's a blocker for us?  thanks

PR #252, opened 2026-08-20
Move the send path onto the claim buffer and rebalancer

guillesd:
Follow-up to #229. That PR built the claim buffer, the rebalancer and the incremental payload writer, but only wired them into the read path; the write path was left on the old design. This PR moves it over and deletes the old one, so both directions now run on the same machinery. Since #229 documents that machinery in detail, this description only covers what is different on the send side.

Net effect is a subtraction: `QuackDataStream` and its registry go away (sequence numbers, per-batch reassembly, dead-range markers, the delivery cursor and its watermark), and `quack_insert.cpp` drops from 478 lines to 130 because the operator becomes a shell over the shared sink, exactly like `QuackFetchCollector` on the server side.

## Wire

SEND_DATA loses `batch_index` (optional), `sequence_index`, `is_last_in_batch`, `batch_watermark` and `dead_range_end`, and gains `ordered` plus `batch_index_fixed`: the same fixed 8-byte patchable field the fetch path uses, through the same generator hooks, so `quack_message.json` stays the wire truth. FINALIZE's `min_batch_watermark` becomes `total_batches`. `QUACK_VERSION` stays at 3, since nothing is released and everything here is still experimental.

The dead-range machinery is gone rather than ported. It existed because executor batch indices are sparse: a predicate that empties whole row groups leaves holes, and the old server cursor would stall on them, so the client had to send explicit "these indices are dead" markers. The rebalancer stamps dense indices, so the stream it produces has no holes by construction and there is nothing to skip.

## Retry safety, now in both directions

Every SEND_DATA names its dense index, so a transport-level retry re-sends the same batch rather than appending a second copy. The claim buffer drops the duplicate, including one that arrives after its batch was already popped, because the `seen` index set is never cleared on pop. FINALIZE then carries the batch count and the buffer closes against it, so a short stream fails loudly instead of quietly inserting less data than the client sent. `quack_debug_duplicate_sends` sends every batch twice on purpose; the test asserts the message count is exactly double and that the insert is unchanged.

On the server, `scan_data_from_quack_client` claims the next dense index and waits for exactly that batch, keeping the claim across a `BLOCKED` yield; unordered streams pop whatever is ready. On the client, `SendDataPayloadWriter` subclasses `QuackChunkPayloadWriter`, so the debug round-trip pin from #229 now guards the send message too.

## TODO: memory management of `QuackInsertStream`

Worth flagging explicitly, because this PR does not fix it and it is the more dangerous half of the memory story.

`SetCapacity` is called in exactly one place, and that is the fetch stream. An insert stream keeps `capacity == 0`, so `PushBatch` never waits. The client's async queue does not help: `ApplyBackpressure` gates on `pending_bytes + in_flight_bytes`, and the server answers a SEND_DATA as soon as it takes the batch, so those bytes leave the client's accounting exactly when they enter the server's buffer. What the server holds is the integral of (arrival rate minus insert rate), with no ceiling except the size of the whole insert. The chunks are allocated from `Allocator::DefaultAllocator()`, so the buffer manager cannot see them, cannot spill them, and `memory_limit` will not stop them. The failure mode is RSS growth to an OOM kill, not a clean error, and it bites hardest on a fast link where nothing throttles the producer.

This is not a regression: the old `QuackDataStream` had no bound of any kind either. It is called out because the symmetry makes the asymmetry obvious. The fetch direction has a byte capacity and its producer parks on it; the send direction has nothing.

The fix is the `accept_budget` field, which is already on the wire and unused, so no protocol change is needed: the server advertises remaining headroom on each SEND_DATA response and the client throttles on it. Two things matter when doing it. The server buffer must keep `capacity == 0` rather than enforcing with `SetCapacity`, because `PushBatch` blocks its caller and the caller is an HTTP request thread. And a client-side credit gate sits in front of the buffer's head-of-stream admission and can defeat it (a producer holding batch N blocked on credit, while other threads have filled the buffer with N+1 and N+2, and the consumer waits for N), so the "one message in flight at budget 0" rule has to come with it, plus a test that starves the lowest index on purpose.

## Follow-ups

- Send-side backpressure via `accept_budget`, as above. Next PR.
- Upstream reserved/backpatchable field support in DuckDB's serializer. #229 flagged this for the fetch writer; there are now two writers depending on the hand-mirrored encoder, so it deletes twice as much.
- Re-measure. The earlier numbers for this design (-22.1% loopback ordered INSERT, -20.5% EC2 to EC2 SF10) predate the rebase onto the #229 machinery and should not be quoted until they are reproduced on the current base.

## Notes for review

`quack_send_data_flush_rows` is deleted; the send path now cuts on `quack_target_batch_bytes` like the fetch path, so batches are sized by serialized bytes rather than row count. Two debug settings are added for the tests: `quack_debug_send_delay_ms` randomises send order to scramble arrival on the server, and `quack_debug_duplicate_sends` sends every batch twice.

The emitter keeps the existing queue-based backpressure (`ApplyBackpressure` holds the producer thread) rather than parking the task the way the fetch emitter does, so `TryEmitPrepared` always returns true here. That is safe because the rebalancer calls the emitter outside its lock. Converging the two is left for the backpressure PR.


src/include/quack_insert_stream.hpp
```diff
@@ -0,0 +1,66 @@
...
+	bool ordered;
+	QuackChunkClaimBuffer buffer;
+};
+
+//! Maps a stream id to its stream. It is process-global, because the scan has a different
+//! ClientContext than the request handler.
+class QuackInsertStreamRegistry {
+public:
+	static QuackInsertStreamRegistry &Get() {
+		static QuackInsertStreamRegistry instance;
```
Mytherin: I'm not quite following why we need this to be a global - can't we store this in e.g. the `DatabaseInstance` then? Or is there a reason for this to be available also across different DuckDB database instances?

As a general rule process-globals are almost always the wrong design for DuckDB and we'd need a very good motivation to use them.

src/include/quack_message.hpp
```diff
@@ -463,17 +444,13 @@ class SendDataRequestMessage : public QuackMessage {
 private:
 	string schema_name;
 	string table_name;
+	//! The chunks of the batch, after the decode.
```
Mytherin: Do we still need the schema / table name here? Maybe this is for a follow-up but my thoughts were that we would try to push these chunks as a scan in a query so that this data transfer method can be used for arbitrary queries (also delete, merge into, etc) and is not only limited to inserting data into a table.

Mytherin (commented):
Thanks for the PR! Looks good - left some comments

src/include/quack_message.hpp
guillesd: the `quack_scan_from_client` went in to a previous PR. So the insert query is already being triggered (currently by a background thread, which again we may need to discuss how to remove it or bound it to a pool) as a scan from a client in server side. This is in `quack_server.cpp:262`:

```cpp
//! Background thread: runs the INSERT that drains `stream` via scan_data_from_quack_client, holding
//! the statement lock for the statement's duration (one transactional statement -> atomic).
static void RunInsertStatement(QuackConnection &connection, shared_ptr<QuackInsertStream> stream, string stream_id,
                               string schema_name, string table_name) {
	try {
		unique_lock<mutex> lock(connection.statement_lock);
		auto sql = StringUtil::Format("INSERT INTO %s.%s SELECT * FROM scan_data_from_quack_client(%s)",
		                              SQLIdentifier(schema_name), SQLIdentifier(table_name), SQLString(stream_id));
		auto result = connection.duckdb_connection->Query(sql);
		if (result->HasError()) {
			stream->buffer.SetError(result->GetErrorObject());
		}
	} catch (std::exception &ex) {
		stream->buffer.SetError(ErrorData(ex));
	}
	// Make sure the consumer side is released even on an unexpected early return.
	stream->buffer.Finish();
}
```

I think your concern is basically that this `INSERT` query is cooked in the server. Yes, we don't need this, it should be driven by the client. I can rework it in this PR if you want, otherwise it can be a follow up. Then we definitely don't need table_name and schema_name.

Mytherin: Did another scan of the code and I think the issue is that currently the `SendDataRequestMessage` is doing too much - it's both (1) starting an `INSERT` query potentially, and (2) sending over chunks.

I think the way I would structure this is more like this:

```sql
CLIENT      SERVER


    START INSERT
      ->
    
    WAITING FOR CHUNKS
     <-


    SEND CHUNK #1
       ->
    ACK
       <-

    ...

    SEND CHUNK #n
       ->
    ACK
       <-
```

i.e. the client should be the one that schedules the `INSERT` query (or any other query) up front, followed by sending chunks using `SendDataRequestMessage`.  `SendDataRequestMessage ` would then be used **only** to send a chunk for a call to `scan_data_from_quack_client`. `<stream id>` should be a cryptographically secure identifier so that we don't need to run authorization for every `SendDataRequestMessage`. 

I think this is both much more flexible (as the client then dictates how / where `scan_data_from_quack_client` is triggered, could even have multiple `scan_data_from_quack_client` in one query, etc) and simpler (as `SendDataRequestMessage` is just sending chunks, instead of also scheduling queries, etc).

Mytherin: This kind of relates to what we discussed before as well where it would be nice if `PrepareRequestMessage` could return something earlier than a result set, e.g. it could also return `BLOCKED` or similar to the client. Then this would just be a `PrepareRequestMessage`:

```
CLIENT      SERVER


    PrepareRequestMessage
    INSERT INTO tbl SELECT * FROM scan_data_from_quack_client(<stream_id>);
      ->
    
    BLOCKED
     <-


    SendDataRequestMessage #1
       ->
    ACK
       <-

    ...

    SendDataRequestMessage #n (final = true)
       ->
    ACK
       <-

    FetchDataRequestMessage #query
       ->
   ROWS CHANGED
       <-

```

src/include/quack_insert_stream.hpp
guillesd: The registry is made so that the httplib handler can register the `{stream_id: payload}` and then it can be picked up by the query by referencing the id ` INSERT INTO s.t SELECT * FROM scan_data_from_quack_client('<stream_id>')`. This was my bad because the `QuackStorageExtensionInfo` at the `DatabaseInstance` level is the place for it! Moved it

src/include/quack_message.hpp
Mytherin: Another thing that's not super clear to me is where the result schema (column names / types) for `scan_data_from_quack_client` comes from - I would say the user should supply these, maybe as part of the query? e.g.:

```sql
INSERT INTO tbl SELECT * FROM scan_data_from_quack_client(..., NULL::ROW(x INT, y VARCHAR));
```

guillesd: Okey I like your point regarding the prepare handler initiating the query and the schema being passed from the client (matches the idea of initiating the query in the client anyway). I'm gonna be reworking it.

What do you think about the fact that currently send data has a `FINALIZE` message? (it is useful but this is not symmetrical to the fetch data path).

Mytherin: Cool cool cool.

I would probably integrate the `final_chunk` into the `SendDataRequestMessage` so we don't need to send a separate message, we could just set `final_chunk = true` for the last chunk we send.

A `FinalizeMessage` could still be emulated by sending along a `SendDataRequestMessage` with `final_chunk = true` and no chunks, so we also don't need this to be a separate message type.

guillesd: Okey so I gave it a go!

guillesd:
- Added a PREPARE that runs the `INSERT INTO s.t SELECT * FROM scan_data_from_quack_client('<id>', NULL::STRUCT(...), ordered := ...)` query. This also means that `schema_name/table_name/ordered` can be removed from the protocol.
- The schema is supplied by the client
- We now send a secure, client minted `stream_id`, so no per-message authorization.
- We remove `FINALIZE` and just send an empty `SEND_DATA` with `total_batches`.

I also did a bunch of renaming because it was getting slightly confusing to follow as there was a lot of "fetch" specific terminology. Some examples:
- `QuackFetchStream → QuackResultStream, QuackFetchState → QuackStatementState, RunFetchQuery → DriveQuery, AbortFetchStream → AbortStatement, SyncFetchCache → SyncResultCache, DetachedFetchStream → DetachedResultStream`

src/quack_message.cpp
```diff
@@ -384,6 +403,26 @@ void FetchResponseMessage::ApplyBatchIndexFixed(const string &bytes) {
 	}
 }
 
+string SendDataRequestMessage::EncodeBatchIndexFixed() const {
```
Mytherin: Why does this need to be transformed back and forth from a string? It seems this is just a `uint64_t`?

src/include/quack_storage.hpp
```diff
@@ -32,9 +33,15 @@ class QuackStorageExtensionInfo : public StorageExtensionInfo {
 	vector<QuackConnectionSnapshot> GetActiveConnectionSnaps();
 	vector<ServerSnapshot> ListServers();
 
+	QuackInsertStreamRegistry &InsertStreams() {
```
Mytherin: I'm still a bit unsure about these being in the storage extension - can we instead have the `SEND_DATA_REQUEST` send over a client id so this can live in the client? That also makes garbage collection simpler (and actually also means we wouldn't need to have a cryptographically secure stream id, as `connection_id` is already cryptographically secure, `connection_id` + stream id as a random uuid or so would be just as safe).

src/quack_server.cpp
```diff
@@ -191,9 +118,10 @@ static void SyncFetchCache(QuackConnection &connection, QuackFetchStream &stream
 	connection.SyncCachedRows();
 }
 
-//! Runs the client's query with the fetch collector installed, and holds the statement lock for
-//! the full duration. The query fills the stream's claim buffer in parallel.
-static void RunFetchQuery(QuackConnection &connection, shared_ptr<QuackFetchStream> stream, string sql) {
+//! Runs one client statement with the result collector installed. It holds the statement lock for the
+//! full duration. A statement that returns rows fills the claim buffer in parallel. A statement that
+//! returns a count leaves the count in batch 1, through the fallback below.
+static void DriveQuery(QuackConnection &connection, shared_ptr<QuackResultStream> stream, string sql) {
```
Mytherin, line 127: If there's a query error I don't think we're aborting the streams correctly (`DropSession` is not called) - this can lead to e.g. a constraint violation not causing an early abort, which might lead the client to still proceed with the full remaining data load despite the query being cancelled. Can we verify this / add a test for this? I think in `SEND_DATA_REQUEST` we should be checking the error message on the query level as well.

src/quack_send_data.cpp
```diff
@@ -0,0 +1,257 @@
...
+		queue->Close();
+		queue.reset();
+
+		auto &quack_catalog = table.catalog.Cast<QuackCatalog>();
+		auto client_wrapper = quack_catalog.GetClientConnection()->GetClient(context);
+		auto &client = client_wrapper->GetClient();
+
+		// The terminal message is an ordinary SEND_DATA with no chunks. It closes the stream against the
+		// batch count, so a lost batch fails the statement.
+		auto terminal = make_uniq<SendDataRequestMessage>(quack_catalog.GetConnectionId(), stream_id,
```
Mytherin: Can we instead set total batches on the final message with chunks we send over if there's any left in the `queue`? That would save us an extra round trip here

Mytherin (commented):
Thanks for the updates - looks great. Some more comments from my side

src/quack_send_data.cpp
guillesd: THis is the same point you had here https://github.com/duckdb/duckdb-quack/pull/229#discussion_r3813320927 and I deferred it to a follow up PR. Maybe that follow up can contain both. If it was very simple I'd do it here, but this requires to again make the `total_batches` field backpatchable so it can be patched to an existing payload in `Finish` when we actually have the real count.

src/quack_message.cpp
guillesd: It is a string only because the current serializer cannot reserve a fixed-width integer, that is what I try to address with https://github.com/duckdb/duckdb/pull/25013 with `EncodePaddedLEB128` (amongst other things). So it is a temporary thing that needs a follow up once some form of the new serializer PR is merged in core

src/include/quack_storage.hpp
guillesd: Done, now on the server-side `ClientContext` of the connection, in `QuackSessionState` (a `ClientContextState`, key `quack_session`).

src/quack_server.cpp
guillesd: Added a test and made sure that `DriveQuery` puts the statement's error on the session's insert streams when the query fails (`Fail`).

guillesd:
@Mytherin I worked on the changes we discussed! 
- Backpatching is gone. I've added the chunk blob with the reserved memory allocation for the header to be prepended.
- I then asked an agent to review the PR and tied some knots

guillesd:
Conflicts resolved. I added the parent schema thing from #261 to the query text that we ship (not needed as a part of the message anymore, same as `table_name` and `schema`)

Mytherin:
Very nice, thanks!

PR #254, opened 2026-08-24
Serve the quack RPC over TLS

Robinbinu:
Lets the server terminate TLS itself, rather than only behind a reverse proxy. Motivated by
duckdb/duckdb#24949.

```sql
CALL quack_serve('quack:localhost:9494', token => '...',
                 tls_cert => '/path/cert.pem', tls_key => '/path/key.pem');
-- listen_url is now https://localhost:9494
```

### Why

The client half already assumes TLS exists. Both entry points default to it for anything non-local:

```cpp
// quack_storage.cpp (QuackAttach) and quack_scan.cpp (bind), identically:
auto enable_ssl = !initial_uri.IsLocal();
```

`QuackUri` has carried an `ssl` flag since the beginning and `quack_serve` already takes `disable_ssl`.
The listener is the one piece that could not do it, and said so:

```cpp
bind_data->listen_uri = QuackUri(listen_uri, /* the server will always listen without SSL */ false);
```

So `ATTACH 'quack:some-host:9494'` reaches for `https://` and finds nothing there, and every remote
deployment has to either front the server with a proxy or turn the default off with `disable_ssl=true`.

### How

`httplib::SSLServer` derives from `Server`, so the bind, the handlers, the elastic task queue and the
listen loop are untouched - only construction differs, plus an `is_valid()` check so a bad certificate
fails at `quack_serve` time with a message about the certificate instead of surfacing later as a bind
error.

`CPPHTTPLIB_OPENSSL_SUPPORT` also switches httplib into its `duckdb_httplib_openssl` namespace, which is
what keeps these translation units from colliding with the plain `duckdb_httplib` that core compiles into
libduckdb. httpfs sets that define per source file; here it has to be the whole extension, because the
`Server` lives in `quack_server.hpp` and every TU that sees the header has to agree on the type. A
`quack_httplib` alias absorbs the namespace difference so the Emscripten build - no OpenSSL, and
`quack_serve` already refuses to run there - still compiles.

### What this does not do

* **Nothing about client-side trust.** With the default curl backend the client already verifies the
  server against the system trust store - `enable_curl_server_cert_verification` defaults to true and
  sets `CURLOPT_SSL_VERIFYPEER` plus `VERIFYHOST=2` - so a real certificate works with no extra
  configuration and a self-signed one needs `ca_cert_file`. Worth knowing, though it is not this PR's
  doing: the optional `httpfs_client_implementation='httplib'` backend takes its default from
  `enable_server_cert_verification`, which is false, so that path encrypts without checking who
  answered. Same URL, two different security properties depending on a client setting.
* **No self-signed certificate generation.** The linked discussion asks for TLS with no cert management
  at all, and `SSLServer` would make it easy. I left it out on purpose: a key baked into the binary is
  public by construction, and one generated per boot still gives encryption without authentication
  unless the client is separately told to trust it. If you want it, an explicit
  `tls_self_signed => true` that refuses non-local binds and prints the fingerprint seems like the
  shape that would not mislead anyone.
* No mTLS. `SSLServer` takes a client CA store and this could grow `tls_client_ca`, left out to keep the
  change small.
* No change to the `allow_other_hostname` rule or its reverse-proxy advice.

### Test

`test/sql/server_tls.test` runs the real thing: serve with the fixture certificate, check `listen_url` is
`https://`, `ATTACH ... (disable_ssl false)`, create, insert, read back. Plus the argument checks - cert
without key, key without cert, and an unreadable certificate.

The fixture in `test/certs/` is a self-signed localhost certificate with its key, valid to 2046. Test key
material in the tree follows `duckdb/test/mbedtls/{public,private}.pem`. Happy to generate it at test time
instead if you would rather not carry it.

`make test` passes.


mrendi29:
Hey @pdet wondering what your thoughts are on this one? IMO this is a critical feature for further enterprise adoption where mTLS is often non-negotiable.

PR #256, opened 2026-08-24
docs+test: pin how the authorization callback's return value is interpreted

guizmaii:
## What

`quack_authorization_function` is documented as `(sid, query) -> BOOLEAN`, but `quack_server.cpp:728` treats a `VARCHAR` return as the SQL to execute, and line 772 runs that instead of the client's query:

```cpp
auto effective_sql = (auth_result.type().id() == LogicalTypeId::VARCHAR)
                     ? auth_result.GetValue<string>()
                     : prepare_request_message.Query();
```

Nothing tests it, so it can disappear in a refactor without anyone noticing.

**No functional change** — a test pinning current behaviour, and a docs section describing it accurately.

## Why the VARCHAR form is worth keeping

`BOOLEAN` can only allow or deny a *whole query*, so it cannot filter **rows**.

That matters for the catalogue queries a client issues on `ATTACH` (`storage/quack_schema.cpp:34`, `storage/quack_table.cpp:72`). Their response carries every schema name and `CREATE TABLE` DDL on the server, so on a server hosting more than one tenant, a client that runs nothing but `ATTACH` learns every other tenant's schemas and table definitions. Returning `false` does not help — it aborts the `ATTACH` and the client gets no connection.

Rewriting those two queries to add a filter is the only way to scope that response today.

## What the test covers

`test/sql/authz_return_value.test` — in each case, what the server does *now*:

- `BOOLEAN` true / false, and `NULL` failing closed
- `VARCHAR` replacing the executed query
- `VARCHAR` returning the query unchanged (passthrough)
- a rewrite narrowing a result the client asked for in full

…and three fail-open behaviours:

- **a `VARCHAR` return can never deny** — the guard checks only `NULL` and a false `BOOLEAN`, so `DROP TABLE t` is answered by the substituted query rather than rejected
- **any other return type allows** — an `INTEGER` return falls through to the client's own query
- **the rewrite does not reach `SEND_DATA_REQUEST`** — line 949 invokes the same callback with a synthesized `INSERT` and applies the deny guard alone, so with a callback returning `SELECT error(...)` an insert through an attached client still lands its rows

I pinned those three rather than changing them, so closing them is a separate deliberate commit. **Happy to write that** if you'd like — making the `VARCHAR` branch apply at 949 and rejecting unexpected return types both seem right to me, but they change behaviour and that's your call.

## Docs

`docs/usage.md` gains a return-value table including the fail-open row, and states plainly that a rewrite cannot deny, that it does not reach the insert path, and that **deciding what to rewrite by matching on raw query text is not a security boundary** — aliases, whitespace and qualified names all defeat it; `json_serialize_sql()` is available inside the callback if the decision matters.

It also notes that a catalogue rewrite must preserve column order, since the client reads those results positionally.

## Note

`docs/usage.md` uses `rpc_*` setting names throughout while the code registers `quack_*` (`src/quack_extension.cpp:149-183`). I followed the file's existing convention rather than mixing the two — happy to fix the file-wide rename here or leave it separate.

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -247,11 +247,57 @@ All settings are regular DuckDB session/global options. Set with
 | Setting                         | Type    | Default                   | Description                                                             |
 |---------------------------------|---------|---------------------------|-------------------------------------------------------------------------|
 | `rpc_authentication_function`   | VARCHAR | `rpc_auth_token`          | Name of a 2-arg scalar function `(sid, token) -> BOOLEAN` used by the server to authenticate clients. |
-| `rpc_authorization_function`    | VARCHAR | `rpc_dummy_authorization` | Name of a 2-arg scalar function `(sid, query) -> BOOLEAN` used by the server to authorize each query. |
+| `rpc_authorization_function`    | VARCHAR | `rpc_dummy_authorization` | Name of a 2-arg scalar function `(sid, query)` used by the server to authorize each query. See [Return values](#authorization-callback-return-values). |
 | `rpc_default_token`             | VARCHAR | *(unset)*                 | Shared secret. The server generates one on `rpc_start`; clients must set it before connecting. |
 
 You can plug in your own auth by creating any scalar function with the
 expected signature and pointing the setting at it.
+
+#### Authorization callback return values
+
+The server interprets the callback's return value as follows
+(`test/sql/authz_return_value.test` pins each case):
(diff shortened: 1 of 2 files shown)


guillesd:
Docs are not really dealt with here anymore. They are in duckdb-web repository. Closing!

Issue #225, opened 2026-07-23
Expire abandoned server-side connections after an idle timeout


ekkuleivonen:
## Problem

Quack removes a server-side connection when a client sends `DISCONNECT_MESSAGE`. If the client crashes or loses connectivity, that message never arrives and the connection remains in `active_connections` indefinitely.

This can retain a DuckDB connection, open transaction, query results, and other session resources. The HTTP keep-alive timeout only closes a transport socket; it does not expire the logical Quack connection.

## Reproduction

On the server:

```sql
CREATE TABLE events(i INTEGER);
CALL quack_serve('quack:localhost', token='test-token');
```

From a second DuckDB process:

```sql
ATTACH 'quack:localhost' AS remote (token 'test-token');
SELECT count(*) FROM remote.main.events;
```

Terminate the client without `DETACH` (for example, with `SIGKILL`). The session remains indefinitely:

```sql
SELECT * FROM quack_active_connections();
```

## Expected behavior

A server-side connection should expire after a configurable period without client activity.

Expiration must:

- never race with an in-flight request;
- remove the connection from `active_connections`;
- release query and connection-local state;
- roll back any transaction owned by that DuckDB connection;
- reject subsequent requests using the expired connection ID.

The timeout can initially be opt-in so this change does not impose a new default lifecycle policy.

## Acceptance criteria

- A client terminated without `DISCONNECT_MESSAGE` eventually disappears from `quack_active_connections()`.
- Requests in progress are not considered idle.
- Connection teardown happens outside the connection-registry lock.
- The expiry worker shuts down cleanly with the server.
- Graceful disconnect continues to remove connections immediately.
- Tests cover expiry, an in-flight request, late requests, and teardown.

Transaction-duration limits, upload quotas, retry/idempotency semantics, and resumable sessions are out of scope.

## Related work

- #80: graceful client disconnect
- #173 / #182: transaction propagation and rollback
- #224: acknowledgements and reconnect support
- #158: explicit query cancellation
- #223: connection-state locking

PR #258, opened 2026-08-26
Add quack_query_timeout to interrupt queries that run past a deadline

### LINKED ISSUE #257: A running query is never cancelled when its client disconnects (server burns CPU indefinitely)
# A running query is never cancelled when its client disconnects (server burns CPU indefinitely)

## Summary

When a client disconnects (crash, `SIGKILL`, network loss) **while a query it started is still executing**, the
server keeps running that query to completion — potentially forever for an unbounded query — with no way to reclaim the
CPU. This is distinct from #225 (idle abandoned *connections*): here the connection is not idle, it is actively
burning cores, and the heartbeat lease never reaps it because the lease is only evaluated when a *new* message arrives
on that connection, which a dead client never sends.

There is also no query-execution timeout of any kind, so even a connected client can pin the server with a single
expensive query.

## Reproduction (observed, DuckDB v1.5.5, quack from current `main`)

Server:

```sql
CREATE TABLE events(i INTEGER);
CALL quack_serve('quack:localhost:9495', token='test-token');
```

Client — run an unbounded query **server-side** and then `SIGKILL` the client process (no `DETACH`, no
`DISCONNECT_MESSAGE`):

```sql
ATTACH 'quack:localhost:9495' AS remote (token 'test-token');
SELECT * FROM quack_query_by_name('remote',
  'WITH RECURSIVE t(n) AS (SELECT 1 UNION ALL SELECT n+1 FROM t) SELECT count(*) FROM t');
```

Measured server-process CPU (`ps -o %cpu`):

| moment | server CPU |
|:--|:--|
| baseline | 0% |
| client running the query | **369%** |
| +15s after `SIGKILL` of the client | 372% |
| +60s (heartbeat-lease boundary) | 372% |
| +90s | 362% |

The server never stops. The 60s heartbeat lease does not help: `RenewConnectionLease` is only called from an incoming
message handler, and no further message arrives for a dead client.

## Where this comes from in the code

- The HTTP handler is synchronous with no disconnect callback (`quack_http_server.cpp`), and the query runs on a
  detached producer thread (`RunFetchQuery`, `quack_server.cpp:196`, spawned at `:772`). Nothing wires socket closure
  to a cancel.
- Lease expiry (`RenewConnectionLease` / `CleanupExpiredConnection`, `quack_server.cpp:320-346`) is only reached
  lazily, from another request on the same connection. There is no background sweep of connections (the only sweep,
  `SweepExpiredCaches`, is inline and handles result caches).
- There is no query-execution timeout setting (`quack_extension.cpp:180-274`).

## Why this is easy to fix — the machinery already exists

The `CANCEL_REQUEST` handler already does a correct cross-thread cancel (`quack_server.cpp:1046-1064`):

```cpp
AbortFetchStream(connection, "Interrupted: query was cancelled");
connection.duckdb_connection->Interrupt();
```

and each connection already records `query_state` (ACTIVE) and `query_started_at`. So both problems can be solved by
**one background reaper thread** owned by the server that periodically scans `active_connections` and fires that same
abort/interrupt pair for any connection whose running query has exceeded a deadline. That single mechanism covers:

1. a **query-execution timeout** (new `quack_query_timeout` setting, default 0 = off), and
2. the **disconnect case** (an orphaned running query is just an ACTIVE query that no one will ever cancel — the reaper
   catches it; optionally paired with a socket-close → `interrupt()` for prompt reclaim).

To avoid interrupting a freshly-started query, the reaper would capture `(query_uuid, query_started_at)` under
`connection.lock` and re-check `query_uuid` is unchanged before interrupting — the same identity check the CANCEL
handler already uses at `quack_server.cpp:1051` — and skip internal catalogue traffic (uuid `{0,0}`).

Note: DuckDB deliberately ships no built-in statement timeout
([duckdb/duckdb#8564](https://github.com/duckdb/duckdb/issues/8564)) and points to `interrupt()` from another thread as
the supported path — which is exactly what the reaper would use, and what Quack's cancel path already does.

## Offer

Happy to implement this as a PR if the maintainers agree on the shape (background reaper vs. per-query timer vs.
socket-close hook). Wanted to align on the approach before writing the concurrency code, given how carefully the
server's locking is structured.

guizmaii:
Closes #257.

## What

Adds a `quack_query_timeout` global setting (seconds, `0` = off, the default) and a per-server background **reaper** that interrupts any SELECT-style query that has run past the deadline.

This addresses both halves of #257:

1. **No query-execution timeout existed** — a single expensive query could pin the server's CPU with no bound.
2. **A disconnected client's running query was never reclaimed** — the heartbeat lease is only re-evaluated when a *fresh* message arrives on the connection, which a dead client never sends. An orphaned running query is just an `ACTIVE` query no one will cancel, so the same reaper catches it.

## How

The reaper reuses the existing `QuackPeriodicWorker` (as the client heartbeat does). Each tick (~1s) it reads `quack_query_timeout`; if non-zero, it snapshots `active_connections` and, for any connection whose `ACTIVE` query started more than the timeout ago, fires the existing abort + `Interrupt()` machinery from the `CANCEL_REQUEST` path.

Design points that keep it correct:

- **Interrupt under `fetch.lock`.** `duckdb_connection->Interrupt()` is connection-global, so it must fire while the reaper can prove the running statement is the query it decided to reap. Holding `fetch.lock` across the `Interrupt()` guarantees that — a new PREPARE cannot install a producer until the lock is released. `SetError` and the producer `join()` run after the lock is released, so the executor reschedule stays off the lock.
- **Identity guard.** The decision captures the query's uuid under `connection.lock`; the abort only fires if the connection's current fetch stream still carries it. Client queries always carry a random uuid, so this is unique per install.
- **`{0,0}` sentinel skipped.** Internal catalogue traffic reuses uuid `{0,0}`, so it is excluded from reaping (reaping a reused sentinel could target the wrong query). Every real client query has a random uuid and is covered.
- **Monotonic deadline.** The deadline is measured off `steady_clock`, so an NTP/DST jump can neither interrupt a young query nor defer the timeout. Wall-clock `query_started_at` is kept for `quack_active_connections` display only.

DuckDB deliberately ships no built-in statement timeout ([duckdb/duckdb#8564](https://github.com/duckdb/duckdb/issues/8564)) and points to `interrupt()` from another thread as the supported path — which is exactly what this uses.

## Scope / trade-offs

- **SELECT-style queries only.** INSERT/`SEND_DATA` streams are not reaped. An abandoned INSERT blocks waiting for client data rather than burning CPU, so it is bounded by the heartbeat-lease reclaim (#225 territory), not this CPU timeout.
- **Granularity ~1s.** A `timeout=1` query can run up to ~2s (deadline + one tick).
- The timeout bounds query-start → final-fetch, i.e. total server-side streaming time, not pure execution.

## Test

`test/sql/query_timeout.test`: off-by-default is a true no-op; an over-deadline query is interrupted and the connection stays reusable; a rapid burst of sub-deadline queries is never spuriously reaped; disabling the setting live takes effect on the next tick. A wall-time assertion confirms the interrupt arrives well under the query's full runtime (proof it was reaped, not merely that an Interrupt surfaced). Adjacent concurrency tests (`cancellations`, `heartbeat_lifecycle`, `heartbeat_config`, `attach`, `fetch_cancel_insert`) still pass.

## Open questions for maintainers

1. `{0,0}` catalogue queries are skipped by the reaper (the exclusion is what makes the uuid guard sound). Fine to leave, or add a generation counter so they can be covered too?
2. The `FINISHED` transition in the PREPARE/FETCH paths isn't under `connection.lock`; a query finishing exactly on a tick boundary can have its result-cache dropped by the reaper (cosmetic, only observable with `quack_enable_reconnects`). Move those writes under the lock, or leave it?
3. Want a socket-close → `interrupt()` hook for *prompt* disconnect reclaim on top of the deadline-based reclaim, here or as a follow-up?

### DIFF
--- a/src/include/quack_server.hpp
+++ b/src/include/quack_server.hpp
@@ -12,6 +12,7 @@
 #include "duckdb/common/shared_ptr.hpp"
 #include "duckdb/common/unordered_map.hpp"
 
+#include "quack_periodic_worker.hpp"
 #include "quack_result_cache.hpp"
 #include "quack_uri.hpp"
 
@@ -130,6 +131,9 @@ struct QuackConnection {
 	//! Read unlocked by the cache sweep and the connection listing, so it must be atomic.
 	atomic<QuackQueryState> query_state {QuackQueryState::IDLE};
 	timestamp_t query_started_at {0};
+	//! Monotonic start time for the query-timeout deadline. Wall-clock query_started_at is for
+	//! display only — a clock jump must not interrupt a running query or defer its timeout.
+	time_point<steady_clock> query_started_steady;
 
 	//! The INSERT this connection is currently driving via a client SEND_DATA stream (one at a time).
 	QuackInsertState insert;
@@ -235,25 +239,23 @@ private:
 	//! Destroys reaped caches on a one-shot detached thread so no response waits on their teardown
 	void DestroyCachesDetached(vector<unique_ptr<QuackResultCache>> doomed);
 
-	//! Background thread: interrupts any query that has run past quack_query_timeout. This also
+	//! One reaper pass: interrupt any query that has run past quack_query_timeout. This also
 	//! reclaims a query whose client has disconnected — the server keeps running it otherwise,
 	//! because the heartbeat lease is only evaluated when a fresh message arrives on the connection.
-	void StartReaper();
-	void StopReaper();
-	void ReaperLoop();
-	//! Interrupt `connection`'s running query if it started more than `deadline_us` before `now`.
-	//! A no-op unless the connection has an ACTIVE query old enough; identity-checked so a query
-	//! started since the deadline was computed is never touched.
-	static void ReapConnectionIfOverdue(QuackConnection &connection, timestamp_t now, int64_t deadline_us);
-
-	std::thread reaper_thread;
-	std::mutex reaper_mutex;
-	std::condition_variable reaper_cv;
-	bool reaper_stop = false;
+	void ReaperTick();
+	//! Interrupt `connection`'s running query if it has run for at least `timeout_seconds`.
+	//! A no-op unless the connection has an ACTIVE, non-sentinel query old enough; identity-checked
+	//! so a query started since the decision is never touched.
+	static void ReapConnectionIfOverdue(QuackConnection &connection, time_point<steady_clock> now,
+	                                    idx_t timeout_seconds);
 
 	string token;
 	//! Per-server random key that seeds the HMAC for client_id_hash.
 	string server_hmac_key;
+
+	//! Enforces quack_query_timeout. Declared last so it is torn down (its dtor joins the thread)
+	//! before active_connections; ~QuackServer also stops it explicitly before any teardown.
+	QuackPeriodicWorker reaper;
 };
 
 class HttpQuackServer : public QuackServer {
--- a/src/include/quack_server.hpp
+++ b/src/include/quack_server.hpp
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <condition_variable>
 #include <queue>
 #include <thread>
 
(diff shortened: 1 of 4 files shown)


Mytherin:
Thanks - but could this not just hook into the existing query timeout mechanisms?

guizmaii:
Thanks @Mytherin! Do you mean `max_execution_time`? I'd missed it — it does abort computational runaways in-process (verified: a recursive CTE / cross-join aborts right at the deadline with an INTERRUPT error).

The gap I ran into is that it's `LOCAL_DEFAULT` scope, so setting it on the server's own connection doesn't reach the connection Quack spawns per served request — a server-side runaway ran unbounded with `max_execution_time` set on the server. Is there an intended way for a server-level setting to propagate to served connections, or would you expect Quack to apply it per served connection? Happy to rework this PR around that if so.

(If you meant a different mechanism, point me at it and I'll use that.)

guizmaii:
@Mytherin Congrats for your acquisition 🙂

PR #260, opened 2026-08-26
Fix data race when snapshotting active connection query state

### LINKED ISSUE #223: Fix data race when snapshotting active connection query state
## Summary

`QuackServer::GetActiveConnectionSnap()` reads mutable per-connection query state under `active_connections_mutex`, while `HandleMessageInternal()` writes the same fields under `QuackConnection::lock`. The different mutexes do not synchronize those accesses, producing a C++ data race.

Affected fields:

- `sql_query` (`std::string`, so concurrent copy/reassignment can access reallocated/freed storage)
- `query_state`
- `query_started_at`

This exists in both the current main and `v1.5-variegata` implementations of `quack_active_connections()`.

## Relevant locking

Writer:

```cpp
std::unique_lock<std::mutex> lock(connection.lock);
connection.sql_query = ...;
connection.query_state = ...;
connection.query_started_at = ...;
```

Reader currently:

```cpp
std::lock_guard<std::mutex> lock(active_connections_mutex);
for (auto &[id, conn] : active_connections) {
    snapshot.sql_query = conn->sql_query;
    snapshot.query_state = conn->query_state;
    snapshot.query_started_at = conn->query_started_at;
}
```

`active_connections_mutex` protects the map/lifetime lookup, not the fields protected by `connection.lock`.

## Suggested fix

Avoid holding both mutex classes at once:

1. Under `active_connections_mutex`, copy the `shared_ptr<QuackConnection>` values into a local vector.
2. Release `active_connections_mutex`.
3. For each retained connection, acquire `conn->lock` and copy its snapshot fields.

The shared pointers preserve lifetime if a connection disconnects between steps, and not nesting the map/per-connection mutexes avoids lock-order inversion.

## Validation

Please add a concurrent test that repeatedly queries `quack_active_connections()` while one or more connections execute/fetch queries. A sequential table-function test cannot detect this race; a TSan run would be ideal.

guizmaii:
Closes #223.

## Problem

`QuackServer::GetActiveConnectionSnap()` read `sql_query`, `query_state` and `query_started_at` under `active_connections_mutex`, while `HandleMessageInternal()` (the PREPARE path) writes those same fields under `QuackConnection::lock`. The two mutexes don't synchronize those accesses, so it is a C++ data race. `sql_query` is a `std::string`, so a concurrent reassignment can free the storage the snapshot is copying from.

## Fix

The exact shape the issue suggested, and the same pattern the server already uses elsewhere for per-connection work:

1. Under `active_connections_mutex`, copy the `shared_ptr<QuackConnection>` handles into a local vector.
2. Release `active_connections_mutex`.
3. For each connection, take its own `lock` and copy the session fields.

Never holds both mutex classes at once, so there is no lock-ordering inversion. `session_id`/`client_id_hash` are set at construction and never rewritten, so they stay outside the lock; `cached_rows` is atomic.

## Test

Existing coverage exercises `quack_active_connections()` under concurrency (`cancellations.test` queries it from one thread while a slow query runs on another); it and the adjacent connection/heartbeat tests still pass.

### DIFF
--- a/src/quack_server.cpp
+++ b/src/quack_server.cpp
@@ -359,15 +359,32 @@ void QuackServer::DestroyCachesDetached(vector<unique_ptr<QuackResultCache>> doo
 }
 
 vector<QuackConnectionSnapshot> QuackServer::GetActiveConnectionSnap() {
+	// The session fields (sql_query, query_state, query_started_at) are written under
+	// QuackConnection::lock, so reading them under active_connections_mutex alone is a data race —
+	// and sql_query is a std::string, so a concurrent write can free the storage we copy. Copy the
+	// connection handles out under active_connections_mutex, release it, then read each connection's
+	// fields under its own lock (never holding both mutex classes at once).
+	vector<shared_ptr<QuackConnection>> connections;
+	{
+		std::lock_guard<std::mutex> lock(active_connections_mutex);
+		connections.reserve(active_connections.size());
+		for (auto &[id, conn] : active_connections) {
+			connections.push_back(conn);
+		}
+	}
 	vector<QuackConnectionSnapshot> result;
-	std::lock_guard<std::mutex> lock(active_connections_mutex);
-	for (auto &[id, conn] : active_connections) {
+	result.reserve(connections.size());
+	for (auto &conn : connections) {
 		QuackConnectionSnapshot snapshot;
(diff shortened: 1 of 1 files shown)


Mytherin:
Thanks!

PR #264, opened 2026-08-30
Fix ATTACH when remote tables have computed DEFAULTs (#132)

### LINKED ISSUE #132: ATTACH fails if there is a table with a field which has a computed DEFAULT value

asubbarao:
Fixes #132.

`ATTACH 'quack://…'` aborts with `Binder Error: Catalog "<alias>" does not exist!` when **any** table in the served database has a DEFAULT that needs catalog resolution (`now()`, `1 + 0`, `nextval(…)`). One such table poisons the ATTACH for the whole catalog.

### Mechanism

`QuackTableSet::LoadEntries` parses each remote `CREATE TABLE` and calls `BindCreateTableInfo`, which rebinds column DEFAULTs client-side. `Binder::BindDefaultValues` builds a binder over `schema.ParentCatalog().GetName()` — the attach alias, which is not registered in the `DatabaseManager` yet, because we are still constructing it. Constant defaults are literals and need no lookup, so they bind; anything requiring function resolution throws.

### Fix

Pass `AlterBindMode::SKIP_BINDING` — the escape hatch `bind_create_table.cpp` already provides. `BindDefaultValues` binds a *copy* (the original is kept for serialization), so the parsed DEFAULT survives on the column and binds on first use, once the catalog exists.

Also included: a real error message when a remote table fails to bind (previously a `CREATE TABLE` failure reported itself as a view-parsing error), naming the catalog, schema and SQL.

### Verified

Against quack `main` + duckdb submodule `f1c54da7` (2026-08-25):

```
INSERT INTO r.t_expr DEFAULT VALUES  ->  x = 1        (NOT NULL satisfied, 1 + 0 applied)
INSERT INTO r.t_ts (v) VALUES (42)   ->  ts NOT NULL  (current_timestamp applied)
```

Full suite: 78 tests, 0 failures. `clang-format`: 0 replacements.

### Why this supersedes #231

#231 stripped defaults with `SetDefaultValue(nullptr)`. That makes the attached table forget the DEFAULT entirely, so `INSERT … DEFAULT VALUES` fails the NOT NULL constraint — it breaks #231's own test at line 52, contradicting that PR's claim that remote `DEFAULT VALUES` still executes. #231 also targeted `v1.5-variegata`; this targets `main`.

### DIFF
--- a/src/storage/quack_table.cpp
+++ b/src/storage/quack_table.cpp
@@ -4,6 +4,7 @@
 #include "quack_scan.hpp"
 #include "duckdb/parser/parser.hpp"
 #include "duckdb/parser/statement/create_statement.hpp"
+#include "duckdb/parser/parsed_data/alter_info.hpp"
 #include "duckdb/planner/parsed_data/bound_create_table_info.hpp"
 #include "duckdb/parser/parsed_data/create_view_info.hpp"
 #include "storage/quack_view.hpp"
@@ -15,8 +16,7 @@ unique_ptr<CreateInfo> ParseCreateTable(const string &sql) {
 	parser.ParseQuery(sql);
 	if (parser.statements.size() != 1 || parser.statements[0]->type != StatementType::CREATE_STATEMENT) {
 		throw BinderException(
-		    "Failed to create view from SQL string - \"%s\" - statement did not contain a single SELECT statement",
-		    sql);
+		    "Failed to parse CREATE TABLE from the remote catalog - \"%s\" - expected a single CREATE statement", sql);
 	}
 	auto &create = parser.statements[0]->Cast<CreateStatement>();
 	return std::move(create.info);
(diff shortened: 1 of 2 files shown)


Mytherin:
Thanks!

PR #266, opened 2026-09-03
added quack version / schema ci step

xproblematique:
Closes https://github.com/duckdblabs/duckdb-internal/issues/9781

PR #165 changed Quack's protocol without bumping `QUACK_VERSION`. This resulted in a `500` error when attaching. 

The current PR adds a CI step to check if `QUACK_VERSION` was bumped when the schema has also changed.

### DIFF
--- a/.github/workflows/MainDistributionPipeline.yml
+++ b/.github/workflows/MainDistributionPipeline.yml
@@ -46,3 +46,23 @@ jobs:
       ci_tools_version: main
       extension_name: quack
       format_checks: 'format'
+
+  check-protocol-version:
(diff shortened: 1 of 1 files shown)


smvv (approved):
thanks!

This CI check helps finding the problem but could also be slow (it fails after pushing code instead of after making changes locally). 

I'm not familiar with quack but i would recommend storing the schema version somewhere once and reference/inject that value into the header (or vice versa). Let me know if I can help with looking into a solution!

Mytherin:
Thanks!

PR #268, opened 2026-09-08
Move from `get-duckdb-version` to `duckdb-submodule-version`

carlopi:
Move from `.github/duckdb-version` file to using the submodule version, via workflow vendored from `extension-ci-tools` (see https://github.com/duckdb/extension-ci-tools/pull/406/changes).

I think this should make so that there is a single place to be bumped, that is the duckdb submodule. Using explicit (additional) commits or tags is also always possible.

This is partially an experiment, but I think this should simplify bumping duckdb in extensions (and the other way around).


Mytherin:
Thanks!

PR #269, opened 2026-09-10
Fix auth callback error being swallowed as a generic Authentication error

guillesd:
This small PR introduces an `AuthCallbackResult` as a generic struct that `EvaluateAuthQuery` returns. It still allows to pass the same `Value` as before but also contains a `error` which can be either the specific Auth/Authz error message or a callback error. In this case, if a callback fails because of a query error (like the parser error that we solved in https://github.com/duckdb/duckdb/pull/25513) we can propagate it as: `Auth/Authz callback failed: Parser error: Transformer for rule 'Statement' returned an unexpected type.`.

Solves https://github.com/duckdb/duckdb/issues/25069


Mytherin:
Thanks!

PR #270, opened 2026-09-10
Correct pushdown docs and document that remote SQL is evaluated in the server's session

rustyconover:
## What

Two documentation-level fixes, plus a note in the code.

**1. `docs/usage.md` documents filter pushdown that isn't enabled.**

The Pushdown section says scans support "both **projection pushdown** and **filter pushdown** (constant comparisons, `IS NULL`, `IS NOT NULL`, `IN`, and `AND`/`OR` combinations)" and that "filters are evaluated server-side", with an `EXPLAIN ... WHERE i = 42` example.

On `main`, `fun.filter_pushdown` and `fun.filter_prune` are commented out in both `QuackScanFunction::GetFunction` and `QuackScanByNameFunction::GetFunction`, the WHERE-building block in `BuildPushdownQuery` is commented out, and the matching `EXPLAIN` assertions in `test/sql/pushdown.test` are commented out as `# disabled for 1.5`. Reworded the section to projection pushdown only.

**2. Client session settings don't reach remote SQL, and that isn't documented.**

`PrepareRequestMessage` carries `sql_query`, `query_uuid` and `inline_rows`, and `QuackServer::CreateNewConnection` opens a plain `Connection(*db)` that inherits the server's settings. So:

```sql
SET integer_division = true;
FROM quack_query('quack:localhost', 'SELECT 7/2');
-- 3.5, bound with the server's integer_division, not the client's
```

Same for `ieee_floating_point_ops` (which turns an error into a value), `default_collation` and `TimeZone`. Added a short section stating this and pointing people at the server.

**3. A note at the two disabled `filter_pushdown` lines.**

This isn't a correctness problem today: the only thing that crosses is a positional projection (`SELECT #1, #3 FROM tbl`), so nothing evaluable travels and no setting can change it. It becomes one when filter pushdown is re-enabled, because filters would cross as SQL text via `filter.ToString(col_name)` and be bound in the server's session. `integer_division` and `default_collation` would then decide what a pushed filter means, and which predicates get pushed is an optimizer decision the user can't see or pin. Left three comment lines at each call site so whoever re-enables them sees it.

## Related

The general version of this is [duckdb/duckdb#25579](https://github.com/duckdb/duckdb/discussions/25579): DuckDB records no metadata for which settings can change a query's result, so anything forwarding session state to a remote executor has to guess at the list.

## Out of scope, but noticed

`docs/usage.md` still uses pre-rename names in most places (`rpc_call`, `rpc_start`, `rpc_default_token`, `rpc_auth_token`, `rpc.call`) alongside newer `quack_*` ones. I left those alone and used the registered name `quack_query` in the example I added. Happy to do the rename separately if it's wanted.


guillesd:
hi @rustyconover! we don't maintain docs for Quack anymore from this repo. They are in duckdb-web!

rustyconover:
@guillesd can you transfer this issue over then, its a button on the right?

